PR #24, opened 2023-06-19
Training

pawelru:
- add training materials (incl. pdf)
- export `get_ref_min` and `get_ref_min_incl_cran`
- set default `config` arg to be a empty list to to avoid re-writting everything if one wants to change one option only
- when querying GH API - use release first and if it is empty then tags
- enhance imports from `cli`
- reorder functions putting the _main_ ones to the top
- rename `install_ip` into `execute_ip` and split it into `solve_ip`, `download_ip`, `install_ip`, `check_ip` to better follow the logic (those are all private funs)
- enhance docs (especially about configuration)


inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+# Motivation
+
+```.default
+Error: object ‘foo’ is not exported by 'namespace:bar'
+```
+
+<br>
+
+```.default
+`foo()` was deprecated in bar 1.0.0.
```
gmbecker, line 41: this isn't going to be solved by a minimum dependency, and as I mentioned in the PI planning, maximum version dependencies might be technically supported by the DESCRIPTION file spec (honestly the writing r extensions manual just mentions "comparison operator" but the only example they give is <= so I'm skeptical).

Furthermore, even if it technically works in R, I'm very doubtful it would be accepted on CRAN.

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+
+:::{.callout-important}
+Do not confuse `verdepcheck` with `revdepcheck` for reverse dependency checks. Those are totally different tools!
+:::
+
+#
+
+Let's call a "different set of dependencies" a _strategy_. Currently there are three _strategies_ implemented:
+
+- **max** - use _development_ version of dependencies
```
gmbecker, line 74: are min and release really a searches like min is, or are these simply  development and prod cohort check? if so, it doesn't doesn't seem like they belong in the same discussion as min.

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+- **release** - use _released_ version of dependencies (e.g. from CRAN)
+- **min** - use _minimal_ version of dependencies as per `DESCRIPTION` file
+
+This tool is looking into a **direct** dependencies only.
+
+## Explanation for direct dependencies
+
+The initial prototype of `min` strategy included recursive find of minimal dependency. This oftentimes lead into usage of already archived package or not able to compile very old package on modern systems architecture.
```
gmbecker, line 82: This is a function of the system you're running your tests on, though, and not an inherent "code failure" in the sense that min is searching for. 

That said we should be using versions of recursive dependencies that were concurrent with the versions of the direct deps being tested. switchr has code to do this. You'll need to transform a switcher manifest to a pak manifest, but that should be easily doable. 

Testing old versions of direct dependencies with modern versions of their recursive dependencies is not a useful or realistic check, in my opinion.  Easy to get breakages just from that mismatch that dont' reflect any real cohort that anyone is actually going to have in their renv/locked site library

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+Therefore only **direct** dependencies are being looked into.
+
+# What are the benefits?
+
+* It forces you to specify minimal version of package dependencies to prevent incorrect package usage.
+* Incorporate information about upcoming breaking changes.
+
+# What's needed?
+
+For **max** strategy the algorithm will look for dependent packages references in a new section `Config/Needs/verdepcheck` of the `DESCRIPTION` file.
```
gmbecker, line 95: in my opinion, this should be an argument, but thats probably because I designed switchr that way. 

If the manifest is something you pass the testing machinery, you can easily tell it to test new branches, etc, without *editing the DESCRIPTION file* which I hate the very concept of doing when the package has not changed. It changes the MD5, which means the package is different, even when its not different, and then you have package versions which have the same version number and different MD5s running around and its a huge awful  mess that didn't need to happen.

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+  - utf8
+  - vctrs
+  - withr
+```
+
+## Discussion:
+
+- One might argue that `foo.package` is fully compatible with the `tibble@1.0.0` syntax but the truth is that `tibble@1.0.0` cannot be installed alongside `dplyr@1.0.0` hence `foo.package` cannot be used with both `dplyr@1.0.0` and `tibble@1.0.0` loaded. Therefore `tibble (>= 1.0.0)` really indicates `tibble (>= 2.1.3)` in this case.
+
+- Because of above, the minimal version specification of dependencies is the _true_ **installable** and **compatible** set of dependencies.
```
gmbecker, line 402: Disagree here. Conceptually, `foo.package`  only needs tibble > 1.0.0 and dplyr 1.0.0. It does not care that dplyr 1.0.0 requires tibble 2.1.3. Thats dplyr's problem, and that requirement is already encapsulated and enforced in dplyr 1.0.0

Look at it this way. The point of this is to figure out if our packge is usable within a particular renv'ed environment, right. that renv has tibble 1.0.0, it isn't going to have dplyr 1.0.0, so we already know the answer is no. We don't get any extra information from also requiring tibble 2.1.3

Don't get me wrong, I'm a big fan of installable cohorts. I have literally written a paper and given numerous research talks about it. It was central in my promotion talk to scientist at genentech. _**BUT**_, that isn't the purpose of htis tool. No one is going to come along and say "man I really want to install the exact minimum requirements to use rlistings 0.5.3, _but no later because I hate all package versions that have come out in the last 5 years!!!_

This tool is for finding requirements, and in that case being discussed here, the only meaningful versioned requirement is dplyr 1.0.0. So long as you ahve that, you're guaranteed to have a version of tibble that works too, so you're good and nothing more need be stated.

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+
+- In order to enter GHA debug lines, `verdepcheck` will continue on error entering `R CMD CHECK` step when e.g. dependency resolve failed. If you encounter `vignette builder 'knitr' not found` error - most likely you failed on resolve or install.
+
+# A few caveats:
+
+None of the test result is _stable_:
+
+  - **min**
+
+    Indirect dependencies are installed as usual using the newest available. Therefore: it is not guaranteed that the new release of indirect dependency will be fully compatible with some old version of direct dependency.
```
gmbecker, line 420: This honestly is a pretty fatal flaw in the paradigm, as I mentioned above, but it is one that you can get around with some work that is already done in switchr.

inst/training_202307/202307.qmd
```diff
@@ -0,0 +1,448 @@
...
+
+    e.g. future version of `tibble` won't be fully compatible with `dplyr@1.0.0`
+
+  - **release**
+
+    The same as above though it's relatively unlikely if you use CRAN-only dependencies as this is checked by CRAN.
+
+  - **max**
+
+    This is the most instable environment changed by pushes to any of the dependencies. Also: the main branch is usually not guaranteed to be correct and you might pull broken package version.
```
gmbecker, line 430: This is why GRAN only builds on version change. last version change is a bit more work (though switchr already knows how, so you can get it from that), but it's _**much much safer**_ than HEAD of main

gmbecker (commented):
Overall seems like a very cool and useful tool, and probably was pretty fun to put together.

I do think there are some things that we may want to rework though, please see my comments

inst/training_202307/202307.qmd
pawelru: Good point. Both release and max don't include version search similar to the one that exists in the min strategy. If it makes it easier to understand the scope I can try do differentiate it more.
I have to admit that at the very beginning I was having a little difficulties with how to conceptually combine those three functionalities under a single product. Both release and max (as we name it right now), even if they are relatively simpler, still give some valuable insights so wanted to keep them here. The common ground here is a "set of dependencies" and my idea was then to call it a _strategy_. Then IMHO those three algorithms fits nicely into that definition.

inst/training_202307/202307.qmd
pawelru: > This is a function of the system you're running your tests on, though, and not an inherent "code failure" in the sense that min is searching for.

True. I tried to include that in the discussions down below about min deps putthing this as it's _installable_ set of dependencies - i.e. not necessarily code-wise minimal set of dependencies. I will try to rephrase that section again.

> That said we should be using versions of recursive dependencies that were concurrent with the versions of the direct deps being tested. switchr has code to do this.

That's a very interesting idea - thanks for putting this. I haven't thought about that this way. I will have to explore it. Maybe we can make a new strategy out of it? Just thinking... I am going to put this in the future work not to miss that.
The problem that I am seeing right now is that we would have to resolve (recursively) the whole dependency tree on our end. It's of course doable but from the very beginning I wanted to keep it as a somewhat thin wrapper on `pkgdepends` and relay on it when it comes to deps resolve, download, install and all of that. Putting custom steps into its workflow it's not an easy task and require so ugly workarounds (something that I did once here but I am not very proud of it). Having this fully custom also comes with a price of maintaining same interfaces and not incorporate all the edge cases that folks have already included. I will have to think about implementation but I agree that feels more _correct_ in the sense of probability

> Testing old versions of direct dependencies with modern versions of their recursive dependencies is not a useful or realistic check, in my opinion

Yes and no. 
- Yes because it's very unlikely to have direct dependency on pkg of >10 years old alongside indirect dependency of few months old. 
- No because it's technically possible hence some safeguards are required. 
- No because it's quite likely if we are thinking about relatively new packages with many releases, e.g. in our field: citril imports chevron of version (newest - 3 releases) and we want to use the newest rtables. I can definitely imagine this scenario.

I am going to create a new slide to discuss it.

inst/training_202307/202307.qmd
pawelru: I actually don't think it is going to be changed that frequently. A typical use that I am for-seeing would be that if once specified it would remain unchanged unless a new dependency is added / removed which will anyway involve changes of the `DESCRIPTION` file.
Having said that, `Config/Needs/*` is generally recommended place to store such metadata for CI that is also used by others.

inst/training_202307/202307.qmd
pawelru: I was expecting this comment :) It is probably the weakest side and encapsulation and who is responsible of what resonates with me very well.

It all stems from the limitations of `pkgdepends` and its resolve dependency algorithm - or to be more fair: our specific requirements.

Please have a look at the example up above - let me re-paste it again here:
```{r}
x <- pkgdepends::new_pkg_deps(c("dplyr@1.0.0", "tibble@1.0.0"))
x$solve()
x$get_solution()
```

```.default
<pkg_solution>
+ result: FAILED
+ refs:
  - dplyr@1.0.0
  - tibble@1.0.0
+ constraints (93):
  (...)
x failures:
* dplyr@1.0.0: Can't install dependency tibble (>= 2.1.3)
* tibble@1.0.0: Conflicts with tibble@1.0.0
* tibble: Conflicts with tibble@1.0.0
```

This pair of packages are not installable together as there is an obvious conflict that `tibble@1.0.0` does not meet version criteria within `dplyr`. It's a significant limitation that we have to bear in mind if we want to keep it as a thin layer on `pkgdepends`.

I have spent considerable amount of time of how to overcome this (and also fit nicely within the whole framework) with no success :( I can't really ask package maintainers for it as I am aware that this is a very specific requirement. This is true that currently named "min" deps is not a code-compabilitywise set of minimal dependencies - it's more of an installable (not to mention compileable) set of packages that is a bit more strict. I will emphasise this once more in the slides.
My view on this is that even if it's imperfect - this is a big step forward. I totally got your point about no-one asking for a specific very old version of a package. This is very unlikely to happen. I was thinking more about set of interdependent packages that are developed in parallel - something we are doing a lot. Always blindly vbumping deps is probably not the best strategy. On the other hand - keeping it unchanged is not correct either. Wanted to find a good tool that will warn you when vbump is needed.

gmbecker: This is... somewhat true, but backwards. The cascade goes the other way

1. You cannot load your hard dependency (dplyr needs tibble >=2.1.3 and can't get it), so
2. you cannot load your package, because you must load all hard dependencies to load your package
3. you cannot install your package (because the installation process includes loading the package, which would fail even if you slipped past the other checks)

Note how I'm framing this. It has nothing to do with tibble per se. Its the dplyr requirement for our package that failed here. That's important. 

Packages (which have non-base-R dependencies) can **_never_** be considered in isolation. Ever. For anything, in any circumstances. 

But this is a deeper issue. There was no reason to force tibble 2.0.0 here when one of the active dependencies forced a higher version. The fact that that happened is a sign something in the algorithm isn't doing the right thing.

Put another way, if dplyr 1.0.0 requires 2.1.3 (and has declared as much), there was no reason to even try with tibble 2.0.0 and dplyr 1.0.0. It was literally impossible for that test to be successful a priori. Thus it was impossible for us to get any new information from running it.

pawelru: > But this is a deeper issue. There was no reason to force tibble 2.0.0 here when one of the active dependencies forced a higher version. The fact that that happened is a sign something in the algorithm isn't doing the right thing.

It was just an experiment (not part of any algorihtm) of an environment that reflects 1:1 with what specified by direct dependencies. 

> Put another way, if dplyr 1.0.0 requires 2.1.3 (and has declared as much), there was no reason to even try with tibble 2.0.0 and dplyr 1.0.0. It was literally impossible for that test to be successful a priori. Thus it was impossible for us to get any new information from running it.

Yes exactly. That was the goal for it - to show that such environment is non-fuctional and malformed. This makes me think that such configuration of dependencies is not really testable hence: not correct? (don't know if "correct" is the right term here but I hope you got my point). You can't specify `tibble@2.0.0` + `dplyr@1.0.0` as you are not able to validate its correctness.

Like I said - some adjustment (here: vbump tibble) needs to be done. In the current implementation it just errors. This is well embedded in the dependency resolution algorighm in pkgdepends. I am currently playing with my own resolution logic that uses min for direct and concurrent for indirect deps. It's definitely not super nice as I am relaying on private class fields but it's sort of working. Let me polish it and I will reach out to you again soon for feedback. But to be clear: that not means that I am convinced about correctness of dependencies specifications. It's just a way how to handle incorrect inputs and correct it on-the-fly.

pawelru:
Hi @gmbecker I have pushed a new version that renames "min" into "min_direct" (to emphasise that only direct deps are searched for min version and indirect are installed as usual) plus also add two more strategies:
- min_cohort - find maximum date of directly dependent packages release dates and use that as PPM snapshot date for dependency resolve. This will limit the versions of indirect dependencies by date of parent package release date.
- min_cohorts - for each direct dependency: find its release date and use it as PPM snapshot for that dependency resolve. Next, combine all the individual resolve outputs and resolve it again. This will limit the versions of indirect dependencies by date of parent package release date.

Both of them should be "stable" over time - unless it has a "@*release" dynamic pointer in any of tested package dependencies.

I believe is that the "min_cohort" strategy is what you had in your mind. When testing it, I discovered a relatively big limitation - when there are many interdependent packages developed in parallel and if one of them needs a recent (e.g. development) version of another then the rest of dependency specification of that package is basically disregarded and the newest versions are used. Therefore we are not testing against min versions specified in DESC but rather against env determined by the most recent dependency out of all your dependencies.
Since it's a common use-case for us, I was thinking how to overcome this and I came up with some middle ground strategy called "min_cohorts" - a similar algorithm but repeated for each dependency separately (and then combine step at the very end that takes maximum for duplicates). I believe that this is much closer to what we want but still theoretically it might happen that your direct dependent package won't work correctly with the newer version of indirect dependent package that came with another directly dependent package.

I am happy to hear your thoughts about that. We can even meet and discuss - that would be faster. Please ping me if you are ready.

pawelru:
Please also find intermediate result created for `teal` repo with a few modifications to pkg deps (doesn't really matter here). Please have a look at the dependencies versions.

<details>
  <summary>min_cohort</summary>

```
r$> x$draw()
deps::/var/folders/m1/hrz0h_ls7gz57rc80tnj41t80000gp/T//RtmpvamTQG/file11dbc75078157 0.13.0.
├─shiny 1.7.4 ✨
│ ├─httpuv 1.6.11 ✨
│ │ ├─Rcpp 1.0.10 ✨
│ │ ├─R6 2.5.1 ✨
│ │ ├─promises 1.2.0.1 ✨
│ │ │ ├─R6
│ │ │ ├─Rcpp
│ │ │ ├─later 1.3.1 ✨
│ │ │ │ ├─Rcpp
│ │ │ │ └─rlang 1.1.1 ✨
│ │ │ ├─rlang
│ │ │ └─magrittr 2.0.3 ✨
│ │ └─later
│ ├─mime 0.12 ✨
│ ├─jsonlite 1.8.5 ✨
│ ├─xtable 1.8-4 ✨
│ ├─fontawesome 0.5.1 ✨
│ │ ├─rlang
│ │ └─htmltools 0.5.5 ✨
│ │   ├─digest 0.6.31 ✨
│ │   ├─base64enc 0.1-3 ✨
│ │   ├─rlang
│ │   ├─fastmap 1.1.1 ✨
│ │   └─ellipsis 0.3.2 ✨
│ │     └─rlang
│ ├─htmltools
│ ├─R6
│ ├─sourcetools 0.1.7-1 ✨
│ ├─later
│ ├─promises
│ ├─crayon 1.5.2 ✨
│ ├─rlang
│ ├─fastmap
│ ├─withr 2.5.0 ✨
│ ├─commonmark 1.9.0 ✨
│ ├─glue 1.6.2 ✨
│ ├─bslib 0.5.0 ✨
│ │ ├─base64enc
│ │ ├─cachem 1.0.8 ✨
│ │ │ ├─rlang
│ │ │ └─fastmap
│ │ ├─htmltools
│ │ ├─jquerylib 0.1.4 ✨
│ │ │ └─htmltools
│ │ ├─jsonlite
│ │ ├─memoise 2.0.1 ✨
│ │ │ ├─rlang
│ │ │ └─cachem
│ │ ├─mime
│ │ ├─rlang
│ │ └─sass 0.4.6 ✨
│ │   ├─fs 1.6.2 ✨
│ │   ├─rlang
│ │   ├─htmltools
│ │   ├─R6
│ │   └─rappdirs 0.3.3 ✨
│ ├─cachem
│ ├─ellipsis
│ └─lifecycle 1.0.3 ✨
│   ├─cli 3.6.1 ✨
│   ├─glue
│   └─rlang
├─teal.data=insightsengineering/teal.data@v0.1.2 0.1.2 ✨👷🏻‍♂️🔧
│ ├─shiny
│ ├─checkmate 2.2.0 ✨
│ │ └─backports 1.4.1 ✨
│ ├─digest
│ ├─formatters 0.5.0 ✨
│ │ ├─checkmate
│ │ └─htmltools
│ ├─lifecycle
│ ├─logger 0.2.2 ✨
│ ├─R6
│ ├─rlang
│ ├─shinyjs 2.1.0 ✨
│ │ ├─digest
│ │ ├─jsonlite
│ │ └─shiny
│ ├─teal.logger=insightsengineering/teal.logger@v0.1.1 0.1.1 ✨👷🏻‍♂️🔧
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─logger
│ │ ├─shiny
│ │ └─withr
│ └─yaml 2.3.7 ✨
├─teal.transform=insightsengineering/teal.transform@v0.2.0 0.2.0 ✨👷🏻‍♂️🔧
│ ├─magrittr
│ ├─checkmate
│ ├─dplyr 1.1.2 ✨
│ │ ├─cli
│ │ ├─generics 0.1.3 ✨
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─magrittr
│ │ ├─pillar 1.9.0 ✨
│ │ │ ├─cli
│ │ │ ├─fansi 1.0.4 ✨
│ │ │ ├─glue
│ │ │ ├─lifecycle
│ │ │ ├─rlang
│ │ │ ├─utf8 1.2.3 ✨
│ │ │ └─vctrs 0.6.3 ✨
│ │ │   ├─cli
│ │ │   ├─glue
│ │ │   ├─lifecycle
│ │ │   └─rlang
│ │ ├─R6
│ │ ├─rlang
│ │ ├─tibble 3.2.1 ✨
│ │ │ ├─fansi
│ │ │ ├─lifecycle
│ │ │ ├─magrittr
│ │ │ ├─pillar
│ │ │ ├─pkgconfig 2.0.3 ✨
│ │ │ ├─rlang
│ │ │ └─vctrs
│ │ ├─tidyselect 1.2.0 ✨
│ │ │ ├─cli
│ │ │ ├─glue
│ │ │ ├─lifecycle
│ │ │ ├─rlang
│ │ │ ├─vctrs
│ │ │ └─withr
│ │ └─vctrs
│ ├─formatters
│ ├─lifecycle
│ ├─logger
│ ├─rlang
│ ├─shiny
│ ├─shinyjs
│ ├─teal.code=insightsengineering/teal.code@v0.2.0 0.2.0 ✨👷🏻‍♂️🔧
│ │ ├─checkmate
│ │ ├─crayon
│ │ ├─lifecycle
│ │ ├─R6
│ │ ├─shiny
│ │ ├─styler 1.10.1 ✨
│ │ │ ├─cli
│ │ │ ├─magrittr
│ │ │ ├─purrr 1.0.1 ✨
│ │ │ │ ├─cli
│ │ │ │ ├─lifecycle
│ │ │ │ ├─magrittr
│ │ │ │ ├─rlang
│ │ │ │ └─vctrs
│ │ │ ├─R.cache 0.16.0 ✨
│ │ │ │ ├─R.methodsS3 1.8.2 ✨
│ │ │ │ ├─R.oo 1.25.0 ✨
│ │ │ │ │ └─R.methodsS3
│ │ │ │ ├─R.utils 2.12.2 ✨
│ │ │ │ │ ├─R.oo
│ │ │ │ │ └─R.methodsS3
│ │ │ │ └─digest
│ │ │ ├─rlang
│ │ │ ├─rprojroot 2.0.3 ✨
│ │ │ ├─vctrs
│ │ │ └─withr
│ │ └─teal.widgets=insightsengineering/teal.widgets@v0.2.0 0.2.0 ✨👷🏻‍♂️🔧
│ │   ├─bslib
│ │   ├─checkmate
│ │   ├─ggplot2 3.4.2 ✨
│ │   │ ├─cli
│ │   │ ├─glue
│ │   │ ├─gtable 0.3.3 ✨
│ │   │ │ ├─cli
│ │   │ │ ├─glue
│ │   │ │ ├─lifecycle
│ │   │ │ └─rlang
│ │   │ ├─isoband 0.2.7 ✨
│ │   │ ├─lifecycle
│ │   │ ├─MASS 7.3-60 
│ │   │ ├─mgcv 1.8-42 
│ │   │ │ ├─nlme 3.1-162 
│ │   │ │ │ └─lattice 0.21-8 
│ │   │ │ └─Matrix 1.5-4.1 
│ │   │ │   └─lattice
│ │   │ ├─rlang
│ │   │ ├─scales 1.2.1 ✨
│ │   │ │ ├─farver 2.1.1 ✨
│ │   │ │ ├─labeling 0.4.2 ✨
│ │   │ │ ├─lifecycle
│ │   │ │ ├─munsell 0.5.0 ✨
│ │   │ │ │ └─colorspace 2.1-0 ✨
│ │   │ │ ├─R6
│ │   │ │ ├─RColorBrewer 1.1-3 ✨
│ │   │ │ ├─rlang
│ │   │ │ └─viridisLite 0.4.2 ✨
│ │   │ ├─tibble
│ │   │ ├─vctrs
│ │   │ └─withr
│ │   ├─htmltools
│ │   ├─lifecycle
│ │   ├─rtables 0.6.1 ✨
│ │   │ ├─formatters
│ │   │ ├─magrittr
│ │   │ └─htmltools
│ │   ├─shiny
│ │   ├─shinyjs
│ │   ├─shinyWidgets 0.7.6 ✨
│ │   │ ├─anytime 0.3.9 ✨
│ │   │ │ └─Rcpp
│ │   │ ├─bslib
│ │   │ ├─sass
│ │   │ ├─shiny
│ │   │ ├─htmltools
│ │   │ ├─jsonlite
│ │   │ └─rlang
│ │   └─styler
│ ├─teal.data=insightsengineering/teal.data@v0.1.2
│ ├─teal.logger=insightsengineering/teal.logger@v0.1.1
│ ├─teal.slice=insightsengineering/teal.slice@v0.2.0 0.2.0 ✨👷🏻‍♂️🔧
│ │ ├─shiny
│ │ ├─checkmate
│ │ ├─digest
│ │ ├─dplyr
│ │ ├─ggplot2
│ │ ├─lifecycle
│ │ ├─logger
│ │ ├─R6
│ │ ├─shinyjs
│ │ ├─shinyWidgets
│ │ ├─teal.data=insightsengineering/teal.data@v0.1.2
│ │ ├─teal.logger=insightsengineering/teal.logger@v0.1.1
│ │ └─teal.widgets=insightsengineering/teal.widgets@v0.2.0
│ ├─teal.widgets=insightsengineering/teal.widgets@v0.2.0
│ ├─tidyr 1.3.0 ✨
│ │ ├─cli
│ │ ├─dplyr
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─magrittr
│ │ ├─purrr
│ │ ├─rlang
│ │ ├─stringr 1.5.0 ✨
│ │ │ ├─cli
│ │ │ ├─glue
│ │ │ ├─lifecycle
│ │ │ ├─magrittr
│ │ │ ├─rlang
│ │ │ ├─stringi 1.7.12 ✨
│ │ │ └─vctrs
│ │ ├─tibble
│ │ ├─tidyselect
│ │ └─vctrs
│ └─tidyselect
├─checkmate
├─lifecycle
├─logger
├─magrittr
├─rlang
├─shinyjs
├─styler
├─teal.code=insightsengineering/teal.code@v0.2.0
├─teal.logger=insightsengineering/teal.logger@v0.1.1
├─teal.reporter=insightsengineering/teal.reporter@v0.1.1 0.1.1 ✨👷🏻‍♂️🔧
│ ├─bslib
│ ├─checkmate
│ ├─knitr 1.43 ✨
│ │ ├─evaluate 0.21 ✨
│ │ ├─highr 0.10 ✨
│ │ │ └─xfun 0.39 ✨
│ │ ├─xfun
│ │ └─yaml
│ ├─lifecycle
│ ├─R6
│ ├─rmarkdown 2.22 ✨
│ │ ├─bslib
│ │ ├─evaluate
│ │ ├─fontawesome
│ │ ├─htmltools
│ │ ├─jquerylib
│ │ ├─jsonlite
│ │ ├─knitr
│ │ ├─stringr
│ │ ├─tinytex 0.45 ✨
│ │ │ └─xfun
│ │ ├─xfun
│ │ └─yaml
│ ├─shiny
│ ├─shinyWidgets
│ ├─yaml
│ └─zip 2.3.0 ✨
├─teal.slice=insightsengineering/teal.slice@v0.2.0
├─teal.widgets=insightsengineering/teal.widgets@v0.2.0
├─bslib
├─covr 3.6.2 ✨
│ ├─digest
│ ├─jsonlite
│ ├─rex 1.2.1 ✨
│ │ └─lazyeval 0.2.2 ✨
│ ├─httr 1.4.6 ✨
│ │ ├─curl 5.0.1 ✨
│ │ ├─jsonlite
│ │ ├─mime
│ │ ├─openssl 2.0.6 ✨
│ │ │ └─askpass 1.1 ✨
│ │ │   └─sys 3.4.2 ✨
│ │ └─R6
│ ├─crayon
│ ├─withr
│ └─yaml
├─dplyr
├─knitr
├─MultiAssayExperiment 1.24.0 ✨
│ ├─SummarizedExperiment 1.28.0 ✨
│ │ ├─MatrixGenerics 1.10.0 ✨
│ │ │ └─matrixStats 1.0.0 ✨
│ │ ├─GenomicRanges 1.50.2 ✨
│ │ │ ├─BiocGenerics 0.44.0 ✨
│ │ │ ├─S4Vectors 0.36.2 ✨
│ │ │ │ └─BiocGenerics
│ │ │ ├─IRanges 2.32.0 ✨
│ │ │ │ ├─BiocGenerics
│ │ │ │ └─S4Vectors
│ │ │ ├─GenomeInfoDb 1.34.9 ✨
│ │ │ │ ├─BiocGenerics
│ │ │ │ ├─S4Vectors
│ │ │ │ ├─IRanges
│ │ │ │ ├─RCurl 1.98-1.12 ✨
│ │ │ │ │ └─bitops 1.0-7 ✨
│ │ │ │ └─GenomeInfoDbData 1.2.9 ✨👷🏻‍♂️
│ │ │ └─XVector 0.38.0 ✨
│ │ │   ├─BiocGenerics
│ │ │   ├─S4Vectors
│ │ │   ├─IRanges
│ │ │   └─zlibbioc 1.44.0 ✨
│ │ ├─Biobase 2.58.0 ✨
│ │ │ └─BiocGenerics
│ │ ├─Matrix
│ │ ├─BiocGenerics
│ │ ├─S4Vectors
│ │ ├─IRanges
│ │ ├─GenomeInfoDb
│ │ └─DelayedArray 0.24.0 ✨
│ │   ├─Matrix
│ │   ├─BiocGenerics
│ │   ├─MatrixGenerics
│ │   ├─S4Vectors
│ │   └─IRanges
│ ├─GenomicRanges
│ ├─BiocBaseUtils 1.0.0 ✨
│ ├─BiocGenerics
│ ├─DelayedArray
│ ├─S4Vectors
│ ├─IRanges
│ ├─Biobase
│ └─tidyr
├─R6
├─scda=insightsengineering/scda@v0.1.5 0.1.5 ✨👷🏻‍♂️🔧
├─scda.2022=insightsengineering/scda.2022@v0.1.3 0.1.3 ✨👷🏻‍♂️🔧
├─shinyvalidate 0.1.2 ✨
│ ├─shiny
│ ├─htmltools
│ ├─rlang
│ └─glue
├─testthat 3.1.9 ✨
│ ├─brio 1.1.3 ✨
│ ├─callr 3.7.3 ✨
│ │ ├─processx 3.8.1 ✨
│ │ │ ├─ps 1.7.5 ✨
│ │ │ └─R6
│ │ └─R6
│ ├─cli
│ ├─desc 1.4.2 ✨
│ │ ├─cli
│ │ ├─R6
│ │ └─rprojroot
│ ├─digest
│ ├─ellipsis
│ ├─evaluate
│ ├─jsonlite
│ ├─lifecycle
│ ├─magrittr
│ ├─pkgload 1.3.2 ✨
│ │ ├─cli
│ │ ├─crayon
│ │ ├─desc
│ │ ├─fs
│ │ ├─glue
│ │ ├─rlang
│ │ ├─rprojroot
│ │ └─withr
│ ├─praise 1.0.0 ✨
│ ├─processx
│ ├─ps
│ ├─R6
│ ├─rlang
│ ├─waldo 0.5.1 ✨
│ │ ├─cli
│ │ ├─diffobj 0.3.5 ✨
│ │ │ └─crayon
│ │ ├─fansi
│ │ ├─glue
│ │ ├─rematch2 2.1.2 ✨
│ │ │ └─tibble
│ │ ├─rlang
│ │ └─tibble
│ └─withr
├─withr
└─yaml
```

</details>

<details>
  <summary>min_cohorts</summary>

```
r$> x$draw()
deps::/var/folders/m1/hrz0h_ls7gz57rc80tnj41t80000gp/T//RtmpxF0P7B/file120042b154b29 0.13.0.
├─shiny 1.6.0 ✨👷🏾
│ ├─httpuv 1.6.6 ✨👷🏾🔧
│ │ ├─Rcpp 1.0.9 ✨👷🏾🔧
│ │ ├─R6 2.5.0 ✨👷🏾
│ │ ├─promises 1.2.0.1 ✨👷🏾🔧 ⬇ (3.12 MB)
│ │ │ ├─R6
│ │ │ ├─Rcpp
│ │ │ ├─later 1.2.0 ✨👷🏾🔧
│ │ │ │ ├─Rcpp
│ │ │ │ └─rlang 1.1.1 ✨👷🏾🔧 ⬇ (762.53 kB)
│ │ │ ├─rlang
│ │ │ └─magrittr 2.0.1 ✨👷🏾🔧
│ │ └─later
│ ├─mime 0.5 ✨👷🏾🔧
│ ├─jsonlite 1.7.2 ✨👷🏾🔧
│ ├─xtable 1.8-2 ✨👷🏾
│ ├─digest 0.6.27 ✨👷🏾🔧
│ ├─htmltools 0.5.1.1 ✨👷🏾🔧
│ │ ├─digest
│ │ ├─base64enc 0.1-3 ✨👷🏾🔧 ⬇ (7.83 kB)
│ │ └─rlang
│ ├─R6
│ ├─sourcetools 0.1.6 ✨👷🏾🔧
│ ├─later
│ ├─promises
│ ├─crayon 1.3.4 ✨👷🏾
│ ├─rlang
│ ├─fastmap 1.1.0 ✨👷🏾🔧
│ ├─withr 2.5.0 ✨👷🏾 ⬇ (102.09 kB)
│ ├─commonmark 1.7 ✨👷🏾🔧
│ ├─glue 1.6.2 ✨👷🏾🔧 ⬇ (106.51 kB)
│ ├─bslib 0.2.5.1 ✨👷🏾
│ │ ├─htmltools
│ │ ├─jsonlite
│ │ ├─sass 0.4.0 ✨👷🏾🔧
│ │ │ ├─fs 1.5.0 ✨👷🏾🔧
│ │ │ ├─rlang
│ │ │ ├─htmltools
│ │ │ ├─R6
│ │ │ └─rappdirs 0.3.1 ✨👷🏾🔧
│ │ ├─jquerylib 0.1.3 ✨👷🏾
│ │ │ └─htmltools
│ │ ├─rlang
│ │ └─magrittr
│ ├─cachem 1.0.5 ✨👷🏾🔧
│ │ ├─rlang
│ │ └─fastmap
│ ├─ellipsis 0.3.1 ✨👷🏾🔧
│ │ └─rlang
│ └─lifecycle 1.0.3 ✨👷🏾 ⬇ (106.85 kB)
│   ├─cli 3.4.1 ✨👷🏾🔧
│   ├─glue
│   └─rlang
├─teal.data=insightsengineering/teal.data@v0.1.2 0.1.2 ✨👷🏾🔧
│ ├─shiny
│ ├─checkmate 2.1.0 ✨👷🏾🔧
│ │ └─backports 1.1.5 ✨👷🏾🔧
│ ├─digest
│ ├─formatters 0.3.2 ✨👷🏾
│ │ └─htmltools
│ ├─lifecycle
│ ├─logger 0.2.0 ✨👷🏾
│ ├─R6
│ ├─rlang
│ ├─shinyjs 2.1.0 ✨👷🏾 ⬇ (1.02 MB)
│ │ ├─digest
│ │ ├─jsonlite
│ │ └─shiny
│ ├─teal.logger=insightsengineering/teal.logger@v0.1.1 0.1.1 ✨👷🏾🔧
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─logger
│ │ ├─shiny
│ │ └─withr
│ └─yaml 2.3.7 ✨👷🏾🔧 ⬇ (94.33 kB)
├─teal.transform=insightsengineering/teal.transform@v0.2.0 0.2.0 ✨👷🏾🔧
│ ├─magrittr
│ ├─checkmate
│ ├─dplyr 1.0.0 ✨👷🏾🔧
│ │ ├─ellipsis
│ │ ├─generics 0.0.2 ✨👷🏾
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─magrittr
│ │ ├─R6
│ │ ├─rlang
│ │ ├─tibble 3.0.1 ✨👷🏾🔧
│ │ │ ├─cli
│ │ │ ├─crayon
│ │ │ ├─ellipsis
│ │ │ ├─fansi 0.4.1 ✨👷🏾🔧
│ │ │ ├─lifecycle
│ │ │ ├─magrittr
│ │ │ ├─pillar 1.4.4 ✨👷🏾
│ │ │ │ ├─cli
│ │ │ │ ├─crayon
│ │ │ │ ├─fansi
│ │ │ │ ├─rlang
│ │ │ │ ├─utf8 1.1.4 ✨👷🏾🔧
│ │ │ │ └─vctrs 0.6.3 ✨👷🏾🔧 ⬇ (967.34 kB)
│ │ │ │   ├─cli
│ │ │ │   ├─glue
│ │ │ │   ├─lifecycle
│ │ │ │   └─rlang
│ │ │ ├─pkgconfig 2.0.3 ✨👷🏾 ⬇ (6.08 kB)
│ │ │ ├─rlang
│ │ │ └─vctrs
│ │ ├─tidyselect 1.2.0 ✨👷🏾 ⬇ (101.50 kB)
│ │ │ ├─cli
│ │ │ ├─glue
│ │ │ ├─lifecycle
│ │ │ ├─rlang
│ │ │ ├─vctrs
│ │ │ └─withr
│ │ └─vctrs
│ ├─formatters
│ ├─lifecycle
│ ├─logger
│ ├─rlang
│ ├─shiny
│ ├─shinyjs
│ ├─teal.code=insightsengineering/teal.code@v0.2.0 0.2.0 ✨👷🏾🔧
│ │ ├─checkmate
│ │ ├─crayon
│ │ ├─lifecycle
│ │ ├─R6
│ │ ├─shiny
│ │ ├─styler 1.0.0 ✨👷🏾
│ │ │ ├─backports
│ │ │ ├─cli
│ │ │ ├─enc 0.1 ✨👷🏾🔧
│ │ │ ├─magrittr
│ │ │ ├─purrr 1.0.1 ✨👷🏾🔧 ⬇ (220.63 kB)
│ │ │ │ ├─cli
│ │ │ │ ├─lifecycle
│ │ │ │ ├─magrittr
│ │ │ │ ├─rlang
│ │ │ │ └─vctrs
│ │ │ ├─rematch2 2.0.1 ✨👷🏾
│ │ │ │ └─tibble
│ │ │ ├─rlang
│ │ │ ├─rprojroot 1.2 ✨👷🏾
│ │ │ │ └─backports
│ │ │ ├─tibble
│ │ │ └─withr
│ │ └─teal.widgets=insightsengineering/teal.widgets@v0.2.0 0.2.0 ✨👷🏾🔧
│ │   ├─bslib
│ │   ├─checkmate
│ │   ├─ggplot2 3.3.6 ✨👷🏾
│ │   │ ├─digest
│ │   │ ├─glue
│ │   │ ├─gtable 0.3.1 ✨👷🏾
│ │   │ ├─isoband 0.2.6 ✨👷🏾🔧
│ │   │ ├─MASS 7.3-60 
│ │   │ ├─mgcv 1.8-42 
│ │   │ │ ├─nlme 3.1-162 
│ │   │ │ │ └─lattice 0.21-8 
│ │   │ │ └─Matrix 1.5-4.1 
│ │   │ │   └─lattice
│ │   │ ├─rlang
│ │   │ ├─scales 1.2.1 ✨👷🏾 ⬇ (270.61 kB)
│ │   │ │ ├─farver 2.1.1 ✨👷🏾🔧 ⬇ (1.27 MB)
│ │   │ │ ├─labeling 0.4.2 ✨👷🏾 ⬇ (10.16 kB)
│ │   │ │ ├─lifecycle
│ │   │ │ ├─munsell 0.5.0 ✨👷🏾 ⬇ (182.65 kB)
│ │   │ │ │ └─colorspace 2.0-3 ✨👷🏾🔧
│ │   │ │ ├─R6
│ │   │ │ ├─RColorBrewer 1.1-3 ✨👷🏾 ⬇ (11.64 kB)
│ │   │ │ ├─rlang
│ │   │ │ └─viridisLite 0.4.1 ✨👷🏾
│ │   │ ├─tibble
│ │   │ └─withr
│ │   ├─htmltools
│ │   ├─lifecycle
│ │   ├─rtables 0.5.1 ✨👷🏾
│ │   │ ├─magrittr
│ │   │ ├─formatters
│ │   │ └─htmltools
│ │   ├─shiny
│ │   ├─shinyjs
│ │   ├─shinyWidgets 0.7.4 ✨👷🏾
│ │   │ ├─anytime 0.3.9 ✨👷🏾🔧 ⬇ (348.90 kB)
│ │   │ │ ├─Rcpp
│ │   │ │ └─BH 1.78.0-0 ✨👷🏾
│ │   │ ├─bslib
│ │   │ ├─sass
│ │   │ ├─shiny
│ │   │ ├─htmltools
│ │   │ ├─jsonlite
│ │   │ └─rlang
│ │   └─styler
│ ├─teal.data=insightsengineering/teal.data@v0.1.2
│ ├─teal.logger=insightsengineering/teal.logger@v0.1.1
│ ├─teal.slice=insightsengineering/teal.slice@v0.2.0 0.2.0 ✨👷🏾🔧
│ │ ├─shiny
│ │ ├─checkmate
│ │ ├─digest
│ │ ├─dplyr
│ │ ├─ggplot2
│ │ ├─lifecycle
│ │ ├─logger
│ │ ├─R6
│ │ ├─shinyjs
│ │ ├─shinyWidgets
│ │ ├─teal.data=insightsengineering/teal.data@v0.1.2
│ │ ├─teal.logger=insightsengineering/teal.logger@v0.1.1
│ │ └─teal.widgets=insightsengineering/teal.widgets@v0.2.0
│ ├─teal.widgets=insightsengineering/teal.widgets@v0.2.0
│ ├─tidyr 1.2.1 ✨👷🏾🔧
│ │ ├─dplyr
│ │ ├─ellipsis
│ │ ├─glue
│ │ ├─lifecycle
│ │ ├─magrittr
│ │ ├─purrr
│ │ ├─rlang
│ │ ├─tibble
│ │ ├─tidyselect
│ │ ├─vctrs
│ │ └─cpp11 0.4.3 ✨👷🏾 ⬇ (304.53 kB)
│ └─tidyselect
├─checkmate
├─lifecycle
├─logger
├─magrittr
├─rlang
├─shinyjs
├─styler
├─teal.code=insightsengineering/teal.code@v0.2.0
├─teal.logger=insightsengineering/teal.logger@v0.1.1
├─teal.reporter=insightsengineering/teal.reporter@v0.1.1 0.1.1 ✨👷🏾🔧
│ ├─bslib
│ ├─checkmate
│ ├─knitr 1.40 ✨👷🏾
│ │ ├─evaluate 0.20 ✨👷🏾
│ │ ├─highr 0.10 ✨👷🏾 ⬇ (15.08 kB)
│ │ │ └─xfun 0.36 ✨👷🏾🔧
│ │ ├─stringr 1.4.1 ✨👷🏾
│ │ │ ├─glue
│ │ │ ├─magrittr
│ │ │ └─stringi 1.7.12 ✨👷🏾🔧 ⬇ (7.60 MB)
│ │ ├─yaml
│ │ └─xfun
│ ├─lifecycle
│ ├─R6
│ ├─rmarkdown 2.17 ✨👷🏾
│ │ ├─bslib
│ │ ├─evaluate
│ │ ├─htmltools
│ │ ├─jquerylib
│ │ ├─jsonlite
│ │ ├─knitr
│ │ ├─stringr
│ │ ├─tinytex 0.42 ✨👷🏾
│ │ │ └─xfun
│ │ ├─xfun
│ │ └─yaml
│ ├─shiny
│ ├─shinyWidgets
│ ├─yaml
│ └─zip 2.2.1 ✨👷🏾🔧
├─teal.slice=insightsengineering/teal.slice@v0.2.0
├─teal.widgets=insightsengineering/teal.widgets@v0.2.0
├─bslib
├─covr 3.0.0 ✨👷🏾🔧
│ ├─jsonlite
│ ├─rex 1.1.1 ✨👷🏾
│ │ ├─magrittr
│ │ └─lazyeval 0.2.0 ✨👷🏾🔧
│ ├─httr 1.3.1 ✨👷🏾
│ │ ├─jsonlite
│ │ ├─mime
│ │ ├─curl 3.0 ✨👷🏾🔧
│ │ ├─openssl 0.9.7 ✨👷🏾🔧
│ │ └─R6
│ ├─crayon
│ └─withr
├─dplyr
├─knitr
├─MultiAssayExperiment 1.24.0 ✨
│ ├─SummarizedExperiment 1.28.0 ✨
│ │ ├─MatrixGenerics 1.10.0 ✨
│ │ │ └─matrixStats 1.0.0 ✨👷🏾🔧 ⬇ (210.65 kB)
│ │ ├─GenomicRanges 1.50.2 ✨
│ │ │ ├─BiocGenerics 0.44.0 ✨
│ │ │ ├─S4Vectors 0.36.2 ✨
│ │ │ │ └─BiocGenerics
│ │ │ ├─IRanges 2.32.0 ✨
│ │ │ │ ├─BiocGenerics
│ │ │ │ └─S4Vectors
│ │ │ ├─GenomeInfoDb 1.34.9 ✨
│ │ │ │ ├─BiocGenerics
│ │ │ │ ├─S4Vectors
│ │ │ │ ├─IRanges
│ │ │ │ ├─RCurl 1.98-1.12 ✨👷🏾🔧 ⬇ (731.48 kB)
│ │ │ │ │ └─bitops 1.0-7 ✨👷🏾🔧 ⬇ (10.81 kB)
│ │ │ │ └─GenomeInfoDbData 1.2.9 ✨👷🏾
│ │ │ └─XVector 0.38.0 ✨
│ │ │   ├─BiocGenerics
│ │ │   ├─S4Vectors
│ │ │   ├─IRanges
│ │ │   └─zlibbioc 1.44.0 ✨
│ │ ├─Biobase 2.58.0 ✨
│ │ │ └─BiocGenerics
│ │ ├─Matrix
│ │ ├─BiocGenerics
│ │ ├─S4Vectors
│ │ ├─IRanges
│ │ ├─GenomeInfoDb
│ │ └─DelayedArray 0.24.0 ✨
│ │   ├─Matrix
│ │   ├─BiocGenerics
│ │   ├─MatrixGenerics
│ │   ├─S4Vectors
│ │   └─IRanges
│ ├─GenomicRanges
│ ├─BiocBaseUtils 1.0.0 ✨
│ ├─BiocGenerics
│ ├─DelayedArray
│ ├─S4Vectors
│ ├─IRanges
│ ├─Biobase
│ └─tidyr
├─R6
├─scda=insightsengineering/scda@v0.1.5 0.1.5 ✨👷🏾🔧
├─scda.2022=insightsengineering/scda.2022@v0.1.3 0.1.3 ✨👷🏾🔧
├─shinyvalidate 0.1.0 ✨👷🏾
│ ├─shiny
│ ├─htmltools
│ ├─rlang
│ └─glue
├─testthat 3.1.5 ✨👷🏾🔧
│ ├─brio 1.1.3 ✨👷🏾🔧 ⬇ (12.63 kB)
│ ├─callr 3.7.2 ✨👷🏾
│ │ ├─processx 3.7.0 ✨👷🏾🔧
│ │ │ ├─ps 1.7.1 ✨👷🏾🔧
│ │ │ └─R6
│ │ └─R6
│ ├─cli
│ ├─desc 1.4.2 ✨👷🏾 ⬇ (79.25 kB)
│ │ ├─cli
│ │ ├─R6
│ │ └─rprojroot
│ ├─digest
│ ├─ellipsis
│ ├─evaluate
│ ├─jsonlite
│ ├─lifecycle
│ ├─magrittr
│ ├─pkgload 1.3.0 ✨👷🏾
│ │ ├─cli
│ │ ├─crayon
│ │ ├─desc
│ │ ├─fs
│ │ ├─glue
│ │ ├─rlang
│ │ ├─rprojroot
│ │ └─withr
│ ├─praise 1.0.0 ✨👷🏾 ⬇ (6.10 kB)
│ ├─processx
│ ├─ps
│ ├─R6
│ ├─rlang
│ ├─waldo 0.4.0 ✨👷🏾
│ │ ├─cli
│ │ ├─diffobj 0.3.5 ✨👷🏾🔧 ⬇ (479.28 kB)
│ │ │ └─crayon
│ │ ├─fansi
│ │ ├─glue
│ │ ├─rematch2
│ │ ├─rlang
│ │ └─tibble
│ └─withr
├─withr
└─yaml
```

</details>

pawelru:
I have pushed a new commit that does the following:
- remove "min_direct" (formerly "min") strategy
- rename "min_cohorts" into "min_isolated" so that it's not confused with "min_cohort".

So the current state is as follows
- max
- release
- min_cohort - find maximum date of directly dependent packages release dates and use that as PPM snapshot date for dependency resolve.
- min_isolated - for each direct dependency: find its release date and use it as PPM snapshot for resolving itself. Next, combine all the individual resolutions and resolve it altogether again.

Both min strategies relies on CRAN PPM snapshots so as to limit the versions of indirect dependencies. The most "correct" strategy is the "min_cohort" where there is only one snapshot used determined by the most recently released direct dependent package. We take all the packages from the whole dependency tree (incl. indirect deps) using the greatest version as of certain date.
The biggest advantage is also its biggest limitation. It's very common that packages are developed in parallel and bumping up dep version to the development version of a package would basically mean to take the rest of dependencies using the greatest version. That makes little value in the context of _minimal_ dependency tests. Moreover, this assumption (that all packages are updated as of given time point) is often not fulfilled and every dependency issues we encountered so far were due to outdated but still completely valid package version. 
Because of the above, we have another strategy "min_isolated" where we resolve dependency tree separately (!) for each dependency using above described logic and then combine resolution and solve it once again (using the greatest version if duplicated). This makes it possible to have more messy environment with direct dependent packages from different dates while ~keeping~ trying to keep indirect deps to be not newer than its parent direct dep. The latter is actually potential point of failure as technically it is possible that indirect dep will be newer than direct dep. That's true when there is an indirect dependency present in more than one directly dependent package dependency tree. Example: let's say a package imports A (last released on 2021/01/01) and B (last released on 2022/01/01). Both A and B imports package C (released on a monthly basis). Isolated resolve of dependency sub-trees results in A and C as of 2021/01/01 plus another subtree with B and C as of 2022/01/01. Combining those two subtrees results in A as of 2021/01/01 that imports C as of 2022/01/01. This is not guaranteed to be fully compatible. CRAN includes such check (i.e. reverse dependency checks) but only against the most recent versions - so it's not guaranteed that it's covered. Plus also we have non-cran dependencies.

PPM-based strategies makes dependency tree to be always resolvable and there is no need to break encapsulation and collapse the whole dependency tree into the tested package dependency specification. That was probably the biggest drawback of the "min_direct" strategy.

Having said that, I think we need to acknowledge these limitations and move forward with what we have now. Of course we can improve as we go.

R/deps_installation_proposal.R
```diff
@@ -9,14 +9,25 @@
...
+#' Next, combine all the individual resolutions and resolve it altogether again.
 #'
-#' Any modification is done for direct (!) dependencies. Indirect ones are installed as usual.
+#' Both "min" strategies relies on PPM snapshot in order to limit the versions of indirect dependencies so that
+#' dependency resolution ends with a package released no earlier than any of its dependency.
+#' However, that's not always true for `min_isolated` strategy - done on purpose.
+#'
+#' Please note that only `min_cohort` and `min_isolated` strategies are "stable". The rest are basing on dynamic
+#' references therefore it results might be different without changes in tested package.
+#' The most straightforward example is `max` strategy in which the environment will be different after any push of
```
gmbecker, line 29: I still think its a  mistake to do an install for non-version-bumping commits in main.

The reason is that when you get a positive result and then tell peoiple "yeah, this works for vx.y.z of my dependency", that isnt' actually true, because you don't have version x.y.z, you have some weird intermediary state that isn't a real version of a package at all.

inst/training_202307/202307.qmd
```diff
@@ -429,16 +484,12 @@ None of the test result is _stable_:
...
-What it is not about:
-
-- This does not perform matrix testing of _all_ dependent package versions.
-- This only looks into package versions - in particular: it does not check R versions or system architectures.
+Therefore, this test should be triggered periodically and not on code change (unless changes includes dependencies). Triggering it on change might give a false message that a failure stems from code changes.
 
 # Future work:
 
 - Idea: add `branch` strategy with recursive resolve for possible replacement of `staged.dependencies`.
+- Idea: `concurrent_min` strategy that uses versions of recursive dependencies that were concurrent (as opposed to max) with the versions of the direct deps being tested.
```
gmbecker, line 492: This should be easy with the ppm_snapshots right? You already have code (I think, by glancing through this) that resolves a direct dependency version to a date. Just use that when grabbing recursives and you should be done.

inst/training_202307/202307.qmd
```diff
@@ -429,16 +484,12 @@ None of the test result is _stable_:
...
 
-# A few caveats:
-
-What it is not about:
-
-- This does not perform matrix testing of _all_ dependent package versions.
-- This only looks into package versions - in particular: it does not check R versions or system architectures.
+Therefore, this test should be triggered periodically and not on code change (unless changes includes dependencies). Triggering it on change might give a false message that a failure stems from code changes.
 
 # Future work:
```
gmbecker, line 490: idea: concept of "min version to check", usage:

1. coalesce on a cohort of direct dependencies. 
2. look in DESC files for those dependencies to see versioned requirements they have  on *their* dependencies
3. set the min versions to even consider for each recursive dep to be the max of all versioned dependencies on it from the cohort from (1), if any.

Should dramatically reduce the number of invalid version checks done in the min_isolated strategy

gmbecker, line 490: Binary search rather than walk through all versions of direct deps during min search.

**even better version**: binary search by dates and just use exact snapshots for those dates. This will get you very close to the edge you're searching for, even if it doesn't exactly find it.

gmbecker: Not only does it get you close to what wha concurrent_min would do, it arguably gives end users of packages something more important, an actual real cran state within which the package would have worked.

A bit of a caveat. could get unlucky with a situation like 

- v 0.1.0 of dep works
- v 0.1.1 breaks our pkg
- v 0.1.2 of dep works

But there are ways of detecting that for slight extra cost

gmbecker (commented):
A few more non-exhaustive comments

inst/training_202307/202307.qmd
gmbecker: ie when your binary search finds an edge, test the versions directly on either side to detect this.

R/deps_installation_proposal.R
pawelru: Technically you are right because package version is imperfect state of code identifier. It would be more correct to use md5 sum or check its source. This can give you unique identifier - as opposed to version. The gist is that this is not well known by users but here all the relevant info is accessible (and then debug-print in the action).

But you have one more interesting comment:

> you have some weird intermediary state that isn't a real version of a package at all

Well that's also true but this is somewhat expected, i.e. to take the development version (even if by definition it could be broken) and try to check your code using it. This needs to be a "soft" / "non obligatory" type of check that should not block merges or anything like that. I hope I made it very clear in docs. The overall goal for this strategy is not to make your code bulletproof but rather notify as soon as possible about upcoming (potentially breaking) changes.

inst/training_202307/202307.qmd
pawelru: Yes exactly! I just named it as "min_cohort" as I personally find it to fit better. But the main idea remained unchanged.

inst/training_202307/202307.qmd
pawelru: I agree. Recursive search should be probably the next step. This will get us closer to actual minimal version environment as well as remove redundant nodes in the dependency tree.

pawelru: Added this to the future work

R/deps_installation_proposal.R
gmbecker: Well that sort of depends on what you think a "Development version of a package" is. Bioconductor, for example, has a formal development repository that only ever has distinctly named package versions in it, ever. GRAN is the same way. 

I think you need to sit down and decide which problem you're trying to solve, because they aren't the same and in fact they ultimately end up being incompatible.

Are you

a) building a tool for CI which tests whether the repo HEADs for multiple packages from SCM work together in an exact, "not real package versions" moment, _**or**_ 
b) building a tool which deals in package versions and informs package dependencies (the versioned dependencies referenced in the pkg name).

Currently, as far as I can tell, some parts of your tool are targeting goal (a), while others (and, I believe, the initial conception of the tool) target goal (b).

Both goals have value. Personally I think goal (b) has *more* value, because while the information it gives is less than that given in (a), that difference is marginal, and said information more correctly fits what users and CRAN maintainers care about anyway.

inst/training_202307/202307.qmd
gmbecker: > Interesting. But note one difference. This algorithm will output you env (which then you can translate into DESC specs) whereas my goal is (at least it was at the very beginning) to take specs from DESC and return true/false. You have described a tool to give you correct answer but here I more look for an answer validator. The intention is that if on false, pkg maintainer will correct it to be true. In particular: it doesn't need to be the "true" min specs - just the specs that doesn't fail. Nevertheless, it's quite interesting. Thanks for sharing. But I see it as an independent product.

I don't understand the reasoning here. Your min plan has to do a search through the possible "environments", as you're calling them, checking them until you get to one that succeeds. 

What I describe above is the same, but a more efficient search which should have fewer failed  checks (substantially so for packages with many versions, O(n) vs O(log n) if we were searching among pkg versions) before returning. It deals in environments only exactly the same way the min version detection algorithm already does. 

The trick here will be to do the fast fourier transform approach. There are thousands of dates you might consider, but we only have to include ones that could be change points, ie ones at which at least one of the reverse dependencies got a new version. That will reduce the thousands of potential dates to a few dozen (low hundreds at most I'd expect), and then the binary search will clobber that down to a few checks total (200 dates would translate to 8 cohort checks, or up to ~16 if you do the double checking for point breakages).

R/deps_installation_proposal.R
pawelru: You are right in your thinking. Historically, I started with b) using GH references only and then realized that a) can be also achieved with no additional input from user side. I rephrased the goal of "min deps testing" into "deps testing" and introduces a _strategy_ concept that help distinguish between those. The truth is that each of the _strategy_ answers different questions as you noticed and should be considered a little bit separate. The common ground is that it all relates to dependencies (actually env created based on deps) as well as I wanted to keep roughly the same input requirements for each of them (new field in DESC). The latter is not really true anymore but it's rather due to simplification (i.e. we don't need it) as opposed to adding new input requirement for a given strategy.
I really tried to explain and emphasise this difference in docs, readme, DESCRIPTION etc. Please let me know if you think it's not enough.

inst/training_202307/202307.qmd
pawelru: > I don't understand the reasoning here. Your min plan has to do a search through the possible "environments", as you're calling them, checking them until you get to one that succeeds.

That is partially true. The main goal for min deps check is to check whether everything works if I use minimal dependencies as per DESCRIPTION file. In particular: I am not looking for (searching) the optimum minimal deps possible. And I am not checking whether min deps in DESC is the optimum one.
Let me put that in the example. Let's say the pkg is fully compatible with all versions of dplyr starting from 1.0.0 but there is "dplyr >= 1.2.3" for some reason. For the check I am thinking of, it's a success as tested pkg does not return any failure. Note that I am not failing because v1.2.3 is different than the _true_ minimal v1.0.0.

Let's then dive into "(...) as per DESCRIPTION file" part of the above sentence. As you correctly pointed out earlier, dependency specifications cannot be really taken as is (to put it simply: replace ">=" with "==") and most likely this would lead to the joint dependency tree conflicts. Please refer to the dplyr >= 1.0.0 together with tibble >= 1.0.0 example discussed earlier. This has to be "corrected" - I can either fail (which means to require inputs that would make joint tree to be resolved) or I can "correct" it on-the-fly. As you correctly pointed out earlier - the former will break encapsulation in a really bad way so I gave a try with the latter and achieve it with PPM snapshot. Now when it comes to the PPM, there is actually no search at all. I take release date for each direct dependent pkgs, calculate max out of that collection and use that date as PPM snapshot. For the min_isloated, it's very similar - take the release date for each directly dependent pkgs, resolve itself (that's the difference) and then combine resolutions and resolve once again. No search / optimization included.

To use mathematical analogy where foo represents a checking function that returns 1 on success and 0 on failure. I can see that you are thinking about this problem as **find** (!) min _x_ for which f(x) > 0. I see it more simpler **as for a given** (!) _x_ check if f(x) > 0. Not so rarely f(x) cannot be calculated and I needed to f(g(x)) > 0. The g function can be debatable but still we have x already given.

Finding an optimal min deps is a definitely interesting question and I very much like your idea how to do that but this does not really belong into "check" role of this package. IMHO it should be a new product.

averissimo (approved):
It looks good!

After this is merged https://github.com/insightsengineering/r-verdepcheck-action/pull/16 should be merged for the action to use the new strategies (and stops using `min`)

PR #26, opened 2023-07-10
Fixes and improvements for strategies

averissimo:
ℹ️ base branch is `training`, not `main`

See https://github.com/insightsengineering/nestdevs-tasks/issues/7 for overall progress

## Notable changes:

* **Adds primary parent reference on each resolved dependency**
  * Useful to debug which package needs to be bumped
  * _TODO_: present this information on the output
* **Move `resolve_ppm_snapshot()` out of `lapply` local definition _(re-used on Rcpp version enforcement)_**
* Shows build error message by `{pkgdepends}`
  * See screenshot below with example when `testthat@3.0.3` fails to compile
* Account for multiple release_dates on `get_release_date()`
  * This fixes problem with `dun** (>= 0.1.3) ` that retrieves tags `0.1.3` and `0.1.3-rc1` 
  * Max date is used
* **Support for github only dependencies** _(for example `dun**` on `chev**` package)_
* Reduce number of calls to `resolve_ppm_snapshots`
* Code cleanup

## How to reproduce via docker container

_(assumes a valid bash and docker/podman is installed)_

```bash
mkdir -p verdepcheck-sandbox/repos
cd verdepcheck-sandbox

# Checkout action
git clone -b new-strategies --depth 1 https://github.com/insightsengineering/r-verdepcheck-action

# Download bash action and `script.R`
curl -o action.bash https://gist.githubusercontent.com/averissimo/17229e87f698626ef8d2463ee80f58a7/raw/c763c2b77d078be2e14d39252f91a5ed61054de7/r-verdepcheck-action.bash
git clone -b verdepcheck_action --depth 1 https://github.com/insightsengineering/random.cdisc.data repos/random.cdisc.data

touch .Renviron # (optional) place `GITHUB_PAT=<your_github_pat>` in the file, although it's not needed
mkdir libs # cache for faster runs after first

# Run docker container
docker run -v ./r-verdepcheck-action:/workspace/r-verdepcheck-action -v ./repos:/workspace/repos -v ./action.bash:/workspace/action.bash -v ./libs:/workspace/libs -v ./.Renviron:/root/.Renviron ghcr.io/insightsengineering/rstudio_4.3.1_bioc_3.17:latest bash ./action.bash -p repos/random.cdisc.data -s min_isolated
```

## Screenshots

image


R/check.R
```diff
@@ -137,46 +136,36 @@ solve_ip.min_isolated_deps_installation_proposal <- function(ip) { # nolint
...
-      i_res
-    }
-  )
+  # Keep primary dependencies versions (in case primary dependency is
+  #  resolved in other dependencies)
+  new_res <- new_res[
+    !(new_res$package %in% deps$package) |
+      new_res$package == new_res$parent
+    ,
+  ]
```
pawelru, line 165: as discussed in our chat conversation, this highlighted part needs to go away
otherwise is a good change

averissimo: Done! :heavy_check_mark:

averissimo:
**[BLUF](https://en.wikipedia.org/wiki/BLUF_(communication))**: When searching for release dates for **non-cran refs** such as `teal (>= 0.10.0)` it returns multiple dates _(from GraphQL GH query)_ and a strategy to select for max date should be safe. Filtering for the actual tag doesn't seem possible/easy.

**Action taken**: Select for max date _(it should be safe, as long as the version is specific `v.#.#.#` with no further "sub-versions" as `v0.10` with return all tags with this prefix `v0.10.0-rc@pre-release`, `v0.10.0` and `v0.10.1`)_

_Context_: GraphQL GH query returns the 2 tags `v0.10.0` and `v0.10.0-rc@pre-release` without any tag information.

_note:_ Teal (>= 0.10.0) is just a public example that replicates a concrete use case.

The reproducible example below shows the problem. If we change the query for an empty string we can see that some tagged commits have Tag information and other don't _(as far I know it's arbitrary with 6 out of 35 having the Tag metadata)_

```GraphQL
{
    repository(owner: "insightsengineering", name: "teal") {
      refs(refPrefix: "refs/tags/", query: "v0.10.0", first: 100) {
        nodes {
          target {
            ... on Commit {
              committedDate
              abbreviatedOid
            }
            ... on Tag {
              abbreviatedOid
              commitResourcePath
              commitUrl
              message
              name
              oid
            }
          }
        }
      }
    }
  }
```

### Teal result 1 _(`query: "v0.10.0"`)_
<details>
<summary>json output</summary>

```json
{
  "data": {
    "repository": {
      "refs": {
        "nodes": [
          {
            "target": {
              "committedDate": "2021-10-08T15:10:35Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-10-08T15:10:35Z"
            }
          }
        ]
      }
    }
  }
} 
```

</details>

### Teal result 2 _(`query: ""` -- empty)_

<details>
<summary>json output</summary>

```json
{
  "data": {
    "repository": {
      "refs": {
        "nodes": [
          {
            "target": {
              "committedDate": "2021-07-01T13:15:34Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-10-08T15:10:35Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-01-27T13:02:30Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-04-07T17:37:33Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-06-09T13:07:34Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-07-29T12:16:11Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-10-14T01:06:31Z"
            }
          },
          {
            "target": {
              "abbreviatedOid": "972e2b3",
              "commitResourcePath": "/insightsengineering/teal/commit/3551e2c36d05f4d569a8385505e6ec367d0d2ef3",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/3551e2c36d05f4d569a8385505e6ec367d0d2ef3",
              "message": "inital version of teal\n",
              "name": "v0.0.1",
              "oid": "972e2b3673a1c8cc0414894ba0e67faa7e37e5e9"
            }
          },
          {
            "target": {
              "abbreviatedOid": "a250362",
              "commitResourcePath": "/insightsengineering/teal/commit/60c961856a446e310b37d2eb002c21a972d2aee4",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/60c961856a446e310b37d2eb002c21a972d2aee4",
              "message": "development version with filtering bug fix\n",
              "name": "v0.0.1.1",
              "oid": "a250362d42cc727b294d62da7fcecab8b52e7385"
            }
          },
          {
            "target": {
              "abbreviatedOid": "7315062",
              "commitResourcePath": "/insightsengineering/teal/commit/4504df76f8c8b1ca841678a1bcd4c84e2a0e7b46",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/4504df76f8c8b1ca841678a1bcd4c84e2a0e7b46",
              "message": "version 0.0.2 that works with the new kadcyla app\n",
              "name": "v0.0.2",
              "oid": "73150626634502d3209445892ea995e51fe1d868"
            }
          },
          {
            "target": {
              "abbreviatedOid": "8fcae40",
              "commitResourcePath": "/insightsengineering/teal/commit/35557f910f6d43baa9b381ee7e7b0735967b1f0e",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/35557f910f6d43baa9b381ee7e7b0735967b1f0e",
              "message": "now with hold_filtering and continue filtering\n",
              "name": "v0.0.2.1",
              "oid": "8fcae40701857fe2aba43ca9e4018a5ff6358bf6"
            }
          },
          {
            "target": {
              "abbreviatedOid": "cfcbe2d",
              "commitResourcePath": "/insightsengineering/teal/commit/080ef53aa67b402ab0ffec8798e7d05e76a934b0",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/080ef53aa67b402ab0ffec8798e7d05e76a934b0",
              "message": "Version 0.0.3\n",
              "name": "v0.0.3",
              "oid": "cfcbe2d37732457842eac812dcc082bd2cff4e8e"
            }
          },
          {
            "target": {
              "committedDate": "2018-06-09T19:16:33Z"
            }
          },
          {
            "target": {
              "committedDate": "2018-11-11T22:02:28Z"
            }
          },
          {
            "target": {
              "committedDate": "2019-02-25T21:00:05Z"
            }
          },
          {
            "target": {
              "committedDate": "2019-07-14T18:57:14Z"
            }
          },
          {
            "target": {
              "committedDate": "2019-10-11T15:40:46Z"
            }
          },
          {
            "target": {
              "committedDate": "2019-10-18T13:03:44Z"
            }
          },
          {
            "target": {
              "committedDate": "2019-12-30T11:36:31Z"
            }
          },
          {
            "target": {
              "committedDate": "2020-04-10T10:08:54Z"
            }
          },
          {
            "target": {
              "committedDate": "2020-06-15T09:24:54Z"
            }
          },
          {
            "target": {
              "committedDate": "2020-07-07T15:31:08Z"
            }
          },
          {
            "target": {
              "abbreviatedOid": "75dc075",
              "commitResourcePath": "/insightsengineering/teal/commit/eb2368d7e2a70e13fb968b0f6dc462811381ac13",
              "commitUrl": "https://github.com/insightsengineering/teal/commit/eb2368d7e2a70e13fb968b0f6dc462811381ac13",
              "message": "* `cdisc_dataset` and `dataset` now return R6 class objects (`RelationalDataset`). * A new `teal_data` function to include datasets and connectors into teal application. * `cdisc_data` function to include datasets and connectors into teal application where a `check` argument still could be used and other consistency tests are performed. * `get_raw_data` can be used to derive raw data from R6 objects e.g. (`RelationalDataset`). * `RawDatasetConnector`, `NamedDatasetConnector` and `RelationalDatasetConnector` to execute custom function call in order to get data from connection. * `CodeClass` to manage reproducibility of the data and relationships between datasets. Not directly exposed to the public interface. * `mutate_dataset` allows to modify dataset or connector via `code` argument or an R script. * `mutate_data` allows to change any dataset in `RelationalData`, `RelationalDataConnector` or `RelationalDataList`. * New wrapper functions to manipulate `RelationalDatasetConnector` and `RelationalDataset` such as `get_dataset`, `load_dataset`, `as_relational`. * New wrapper functions to manipulate `RelationalDataConnector`, `RelationalData` and `RelationalDataList` such as `get_datasets`, `load_datasets`. * `choices_labeled`, `filter_spec`, `select_spec`, `data_extract_spec`, `value_choices`,  `variable_choices` as S3 class applied on `data.frame` and also on delayed data.  * You can no longer modify the `app$datasets`, but must instead use argument `filter` in the `init` function.  * New modules were created to create a module of nested teal modules, then another one that adds the right filter pane to each tab. The `teal::init` function stays unchanged. * The `teal::init` function now returns a `UI` function with an optional `id` argument. This allows to embed it into other applications. A split view of two teal applications side-by-side is one such example and shown in a vignette. `teal::init` was turned into a wrapper function around `module_teal_with_splash.R` and developers that want to embed teal as a Shiny module should directly work with these functions (`ui_teal_with_splash` and `srv_teal_with_splash`) instead of `teal::init`. * The `teal::init` function now has a title parameter to set the title of the browser window.\t * Missing data `NA` is now explicitly addressed in the filter panel: `NA`s are excluded by default and a checkbox to include them was added. * Statistics of the data are visually depicted in terms of histograms or bar charts overlayed onto the Shiny input elements. * Added buttons to remove all filters applied to a dataset. * Restored the functionality to hide the filter panel for a module when it was constructed with `filters = NULL`. * Moved helper functions into `utils.nest` and removed unused functions `set_labels_df` and `get_labels_df`.  * `optionalSelectInput` now allows for grouped choices.  ## Refactor of FilteredData (for developers)  * `FilteredData` is now fully reactive. Now filtered data is lazy evaluated as per need. This further opens the door to bookmarking `teal` apps (bookmarking currently works for the right filtering panel, but we will make this feature more sophisticated in a future release, each module must be reviewed and adapted if it contains `reactiveValues`). * Datasets and materialized connectors are provided to `FilteredData` by `set_datasets_data` function located in `init_datasets.R` file. * Renamed `get_dataset()` method to `get_data()`. * Renamed `get_filter_call()` method to `get_filter_expr()`; returns an expression rather than a list. * Removed argument `isolate` from `get_data()` method and similar methods. You must `isolate` it yourself as needed. If you want to temporarily deactivate Shiny errors due to missing errors, you can set `options(shiny.suppressMissingContextError = TRUE)`. In general, avoid `isolate` as this breaks reactivity. * We added a development module to add several filters at once, e.g. safety filters. This is to be evaluated before it is converted into a proper module and made available to end-users.",
              "name": "v0.9.0",
              "oid": "75dc075b4c396891cbfa0314ac7fffb62476b33d"
            }
          },
          {
            "target": {
              "committedDate": "2020-10-19T12:12:30Z"
            }
          },
          {
            "target": {
              "committedDate": "2020-12-15T13:55:22Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-03-18T17:49:08Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-05-03T07:47:36Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-07-01T13:15:34Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-10-08T15:10:35Z"
            }
          },
          {
            "target": {
              "committedDate": "2021-10-08T15:10:35Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-01-27T13:02:30Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-04-07T17:37:33Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-06-09T13:07:34Z"
            }
          },
          {
            "target": {
              "committedDate": "2022-10-14T01:06:31Z"
            }
          },
          {
            "target": {
              "committedDate": "2023-05-22T18:00:53Z"
            }
          }
        ]
      }
    }
  }
} 
```

</details>

pawelru:
> When searching for release dates for non-cran refs such as teal (>= 0.10.0) it returns multiple dates

Very interesting finding. Checked this on my own and I do confirm.

I looked into the code to see how you addressed it and it seems that you fixed it on the _consumer_ side to handle returned vector of dates. Wouldn't be easier / better to assure that `get_release_date` returns obj of length of one (in particular: `NA` for not found or unknown remote-ref class) - that is: handle it on the _producer_ side? As a general note, number of _consumers_ tends to grow over time and each of them would have to have this "fix" which is not so DRY solution.

R/get_ref.R
```diff
@@ -150,8 +150,19 @@ get_ref_min.remote_ref_github <- function(remote_ref, op = "", op_ver = "") {
...
+    remote_ref$username,
+    remote_ref$repo,
+    ref_suffix
+  )
+  result <- pkgdepends::parse_pkg_ref(new_ref)
+
+  # Needs to restore original ref as it needs it to solve the pkg tree. With error:
+  #  Can't install dependency <original_ref> (>= <version>)
+  result$ref <- remote_ref$ref
+  result
```
pawelru, line 165: I am very interested in reasoning here. Can you please elaborate more?
I am a little concerned because this is going to be the first break of self-consistency of this object which is returned(!) (out of exported! fun) so we don't really know who and how it is used. From what I recall, sometimes we use `ref` element and sometimes we use other like `version` etc. With this change we need to be very very careful. And I would like to avoid such traps if possible

averissimo: The previous behaviour was breaking consistency as it changed the original `ref` from the remotes with the output of this function.

It replaces `insightsengineering/dunlin@*release` with `dunlin=insightsengineering/dunlin@v0.1.3`. _(screenshot below with the error and `get_resolution()`)_

As a result, when it is trying to solve the dependencies it detects that the `...@*release` ref cannot be installed.

An alternate solution would be for us to populate the `version`, `commitish` and any other field fields manually. However, I really like using `pkgdepends::parse_pkg_ref` for this as it reduces possible maintenance in the future.

image

image

averissimo:
> Wouldn't be easier / better to assure that `get_release_date` returns obj of length of one (in particular: `NA` for not found or unknown remote-ref class) - that is: handle it on the _producer_ side? As a general note, number of _consumers_ tends to grow over time and each of them would have to have this "fix" which is not so DRY solution.

Very good point. I was doing this on the consumer side as it would make the `get_ppm_snapshot_by_date()` robust to unexpected parameter lengths _(as it just happened)_

Better to enforce the select strategy on the _producer_! I'll make that change.

Is it too late to include `checkmate` and start adding some parameter assertions to better prevent this?

pawelru:
> Is it too late to include `checkmate` and start adding some parameter assertions to better prevent this?

I've been thinking about that and decided to go more modest way when it comes to deps at least for the beginning until we see a really strong reason to do so. This could be one of the reasons but I don't think it's so strong to enforce it. Let's wait and check if we find more cases where it will provide value.

R/get_ref.R
averissimo: I've been thinking about this, testing and debugging and trying to understand the whole context better and design a good solution.

🗓️ I've sent an invite for us to talk on early next week (Monday?) about this.

#### Addressing your last point first:

> it looks like an installation error as opposed to dependency resolve conflicts

It fails to install as the resolution had no candidate for  the `.../...@*release` as the ref name was replaced by `dunlin=.../dunlin@v0.1.3`.

I'm very confident that it's not a token or installation problem as the message appears to indicate.

#### The forced assignment after `pkgdepends::parse_pkg_ref(new_ref)`

⚠️ The comments bellow are only for the `min_isolated` strategy

I think the problem we have with `Remotes` is a bit wider than we realised, there are currently 3 issues _(maybe more as I found a weird big also on `min_cohort`)_

1. Your PR on https://github.com/r-lib/pkgdepends/pull/321
1. Can't install dependency `.../...@*release`
    * Proposed fix that started this discussion
1. Min version is not respected by packages that are described on Remotes dependency that also exist on CRAN
    * For example, `formatters` _(as a primary dependency)_ will install `0.5.9003` instead of `0.5.0`

To help me understand better the whole process I made a diagram for this strategy's technical workflow _(see at the end)_

### 💡 Idea for solution 

I think the solution should be targetted when the IP is being prepared _(during first call on the diagram)

In particular, when the `Need/Config/verdepcheck` resolves to a particular version or to CRAN, it will modify/remove the corresponding remote.

* If it resolves to a particular tag, it changes the `Remotes` to that version instead of `@*release`. If the package is resolved to CRAN then it will remove the corresponding `Remotes`.
* (alternative) Solve this downstream _(second call on the diagram)_
  * For issue 2. using my first fix
  * For issue 3. implement logic that detects this scenario and uses a similar solution

min_isolated drawio(4)

averissimo: ## Summary:

@pawelru I've Implemented the solution we discussed on our meeting today and my initial tests show it's:

* ✅ working very well with `{chevron}` _(pkg where some of these problems were materializing)_
* ❌ The local edge case we talked about confirms our hypothesis with the encapsulation scenario
    * When a primary dependency is included in the dependency tree with a `Remotes` _(resolving to `main` branch)_

Here are the changes: [6079c3c4..caea5c37](https://github.com/insightsengineering/verdepcheck/pull/26/files/6079c3c439b7ac316da7093d2ef0220cbad1bd5c..caea5c374e923354a0c3baea9336cd5067ca54b1) -- Changes applied to all strategies, but `max`.

### On `chevron` it :

* Resolves problematic package `{dunlin}` without any issues _(installing the expected tag)_
* Corrects another bug _(that we didn't discuss)_ with `{formatters}`, installing the "min_isolated" version instead of the latest from main.

### Local edge case (`{test.repo}`)

_(`Imports: rtables, tern and averissimo/teal.logger`)_

`averissimo/teal.logger` is a non-CRAN pkg with custom `Remotes` that includes `formatters`, which is one of the primary dependencies of `{test.repo}`

It tries to install `formatters 0.5.0.9004` instead of `0.5.0` and it validates our hypothesis with the encapsulation problem.

⚠️ note: all packages that are in CRAN will have GitHub `Remotes` ignored and squashed in favor of "ppm", so this example with only `{tern}` and `{rtables}` doesn't have any issues.

image

image

R/deps_installation_proposal.R
```diff
@@ -299,6 +302,56 @@ get_refs_from_desc <- function(d) {
...
+  remotes_pkg <- vapply(remotes, `[[`, character(1), "package")
+
+  # Find which packages of the new_refs are defined in Remotes
+  new_refs_gh <- vapply(
+    new_refs,
+    function(.x) {
+      isTRUE(.x$package %in% remotes_pkg) && inherits(.x, "remote_ref_github")
+    },
+    logical(1)
+  )
```
pawelru, line 332: here it would be more efficient to make use of `Filter()` as opposed to build a mask and then apply it on the vector.

R/get_ref.R
```diff
@@ -150,8 +150,16 @@ get_ref_min.remote_ref_github <- function(remote_ref, op = "", op_ver = "") {
...
+  new_ref <- sprintf(
+    "%s=%s/%s%s",
+    remote_ref$package,
+    remote_ref$username,
+    remote_ref$repo,
+    ref_suffix
+  )
+  result <- pkgdepends::parse_pkg_ref(new_ref)
+
+  result
```
pawelru, line 162: `result` seems to be not needed

```suggestion
  pkgdepends::parse_pkg_ref(new_ref)
```

R/deps_installation_proposal.R
```diff
@@ -299,6 +302,56 @@ get_refs_from_desc <- function(d) {
...
+  #  resolved to be a github repo
+  d$clear_remotes()
+
+  new_ref_pkg <- vapply(new_refs, `[[`, character(1), "package")
+  new_remotes <- c(
+    # Keep remotes
+    d$get_remotes()[!(remotes_pkg %in% new_ref_pkg)],
+    # Modified remotes
+    new_ref_remote
+  )
```
pawelru, line 347: generally speaking it seems that it's just `new_remotes_refs <- setdiff(remotes_refs, new_refs)` - i.e. remove remotes that exists in the second fun arg. You seem to do `new_remotes_refs <- c(setdiff(remotes_refs, setdiff(new_refs, remotes_refs)), setdiff(new_refs, remotes_refs))` which is a little bit more complex to follow and fully understand. Am I right with my thinking? Is there any reason why we are doing it this way?

R/get_ref.R
```diff
@@ -96,7 +96,14 @@ get_ref_min.remote_ref_cran <- function(remote_ref, op = "", op_ver = "") {
...
   new_ref <- sprintf("%s@%s", remote_ref$ref, names(min_ver)) # @TODO deparse, add ver, parse again
-  pkgdepends::parse_pkg_ref(new_ref)
+  tryCatch(
+    pkgdepends::parse_pkg_ref(new_ref),
+    error = function(err) {
+      cli::cli_alert_danger(
+        "Problem finding version for: `{remote_ref$package} ({op} {op_ver})`"
+      )
+      stop(err)
+    })
```
pawelru, line 106: I would say that `pkgdepends::parse_pkg_ref()` is quite low-risk call hence I don't expect it to fail. Can you give me an example where you have an error, i.e. when tryCatch will be useful?

R/deps_installation_proposal.R
averissimo: `vapply` seems a bit faster, but they're pretty neck in neck.

I do like the clean code of using `Filter`!

```R
Browse[1]> .f <- function(.x) isTRUE(.x$package %in% remotes_pkg) && inherits(.x, "remote_ref_github")

Browse[1]> microbenchmark::microbenchmark(
    times = 100L,
    check = "equal",
    filter = Filter(.f, new_refs),
    vapply = new_refs[vapply(new_refs, .f, logical(1))],
    setup = new_refs[] # to remove variability on first call
  )
#Unit: microseconds
#   expr    min      lq     mean  median      uq    max neval
# filter 49.123 49.7765 51.15354 50.1840 51.0315 83.586   100
# vapply 45.296 46.1205 46.99391 46.6455 47.2345 57.032   100

Browse[1]> microbenchmark::microbenchmark(times = 100L, vapply = new_refs[vapply(new_refs, .f, logical(1))])
#Unit: microseconds
#   expr    min     lq     mean  median      uq    max neval
# vapply 44.908 46.119 49.94768 47.6345 53.4715 89.241   100
  
Browse[1]> microbenchmark::microbenchmark(times = 100L, filter = Filter(.f, new_refs))
#Unit: microseconds
#   expr    min      lq     mean median     uq    max neval
# filter 49.841 50.5585 53.84665 51.382 56.061 133.48   100
```

R/deps_installation_proposal.R
```diff
@@ -299,6 +302,56 @@ get_refs_from_desc <- function(d) {
...
+  new_ref_remote <- vapply(new_refs[new_refs_gh], `[[`, character(1), "ref")
+
+  # Remotes that are not in new_refs are kept, as well as the ones that were
+  #  resolved to be a github repo
+  d$clear_remotes()
+
+  new_ref_pkg <- vapply(new_refs, `[[`, character(1), "package")
+  new_remotes <- c(
+    # Keep remotes
+    d$get_remotes()[!(remotes_pkg %in% new_ref_pkg)],
```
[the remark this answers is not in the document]
averissimo: 🤦 Thanks for noticing this, I'll move the code around

R/deps_installation_proposal.R
averissimo: The idea here is to remove all packages that have been resolved to a CRAN package and modify the ones that resolved to a GitHub tag.

> `new_remotes_refs <- c(setdiff(remotes_refs, setdiff(new_refs, remotes_refs)), setdiff(new_refs, remotes_refs))`

I think the code below better shows the logic _(although `setdiff` cannot be used as we're comparing package name, but use remote reference)_

note: the `setdiff` should only output a non-empty vector if the DESCRIPTION is badly formed

```R
c(
  setdiff(remotes_refs, new_refs), # should always be empty unless DESCRIPTION is badly formed
  intercept(remote_refs, new_refs[is.github.tag])
)
```

Transforming:

```R
# (...)
Imports:
  a,
  b,
  c
Remotes:
  ie/b@*release
  ie/c@*release
Config/Need/verdepcheck:
  ie/a,
  ie/b,
  ie/c
```

To:

```R
# (...)
Imports:
  a (>= 1.0.0),
  b (>= 2.0.0),
  c (>= 3.0.0)
Remotes:
  b=ie/b@v2.0.0
Config/Need/verdepcheck:
  a,
  b=ie/b,
  c # exists on CRAN, this is done by `desc_cond_set_refs()`
```

R/get_ref.R
averissimo: This happens when there's an invalid version for a give package, let's say `tern (>= 10.8.3)`

The error when this happens is show below, which I don't think has a good description.

ps.: I'm changing the message to be `"Possible problem finding release for: {remote_ref$package} ({op} {op_ver})"`. It seems like a better description.

```R
Error in "pkgdepends::parse_pkg_ref(new_ref)" :
! `ref` must be a string (character scalar), but it is an empty character vector.
```

image

R/deps_installation_proposal.R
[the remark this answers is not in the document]
averissimo: :heavy_check_mark:  Done in ac12fea

R/deps_installation_proposal.R
averissimo: :heavy_check_mark:  Done in ac12fea

R/check.R
```diff
@@ -137,46 +136,25 @@ solve_ip.min_isolated_deps_installation_proposal <- function(ip) { # nolint
...
-        config = list(dependencies = "hard", cran_mirror = ppm_repo)
-      )
-      suppressMessages(i_pkg_deps$resolve())
-      i_res <- i_pkg_deps$get_resolution()
-      i_res$direct <- i_res$directpkg <- FALSE
-      i_res
-    }
-  )
+  new_res <- do.call(rbind, deps_res)
+  new_res <- new_res[!duplicated(new_res[, c("ref", "package", "version", "parent")]), ]
```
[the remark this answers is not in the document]
averissimo: The parent is just a _"nice to have"_ for faster resolution of problems. 

The original motivation for this was due to Rcpp-related, as it allowed me to bump packages that needed Rcpp at once, instead of iteratively.

But as you said, it adds unnecessary rows and it may have a negative impact. 

So I guess a removal is in order?

R/deps_installation_proposal.R
```diff
@@ -299,6 +302,59 @@ get_refs_from_desc <- function(d) {
...
+#' Replaces any existing Remotes entry with the resolved GitHub tag from the
+#' `new_refs`.
+#'
+#' It keeps all the existing Remotes that have not been resolved in `new_refs`.
+#'
+#' @param d (`desc`) DESCRIPTION object
+#' @param new_refs (`list`) remote references that have been resolved and are
+#' being updated in `Config/Needs/verdepcheck`
+#' @keywords internal
+desc_remotes_cleanup <- function(d, new_refs) {
```
pawelru, line 317: I still struggle to fully understand what it does and why it is there.

1. What it does?
You have the following:
```
  new_remotes <- c(
    # Keep remotes (if the DESCRIPTION file is correct, this should have no elements)
    d$get_remotes()[!(remotes_pkg %in% new_ref_pkg)],
    # Modified remotes
    new_ref_remote
  )
  ```
  
`d$get_remotes()[!(remotes_pkg %in% new_ref_pkg)]` is a vector of remotes not present in config/needs/verdepcheck
`new_ref_remote` is vector with config/needs/verdepcheck refs for packages existing in remotes field
So essentially it's (conditionally) replacing remotes entries with config/needs/verdepcheck entries. OK seems that I get what it does while writing this comment.
UPDATE: I just came up with example where you actually remove some entries so it's not only about replacing one with another. Then I guess we can modify for loop with `if (cond1) then NULL else if (cond2) x else y`

2. Implementation:
- your function changes the order of remotes entries - don't yet know if it's relevant
- you call `$get_remotes()` twice - also not a big deal
- I was probably struggling because of names of var - there is `new_refs`, `new_ref_remote`, `new_ref_remote` and none of them indicates new references for remote field
How about a simple for loop:
```
lapply(
  pkgdepends::parse_pkg_refs(d$get_remotes()),
  function(x) {
    if (<some condition>) then x else config_needs_verdepcheck[get_pkg_name(config_needs_verdepcheck) == x$package]
  }
)
```
- (connected to the above) Let's call `desc_cond_set_refs` first and then `desc_remotes_cleanup` using one argument only indicating desc object. This could potentially simplify naming issue as well as it fits to the role better - you want to replace desc remotes field with another field (that is being extracted from the same desc object and not another argument)

3. Is it really needed?
Please have a look at `get_refs_from_desc` executed at the very beginning of each installation proposal constructor - it is that function which is responsible for preparing correct list of references for further analysis. I would even ask if we need a Remotes at all - as we already / should have everything we need in the config/needs/verdepcheck. That field can potentially have downstream implications so I am inclined to request to always clear it.

File: R/utils.R [hunk 10]
```diff
@@ -31,3 +34,37 @@ get_ppm_snapshot_by_date <- function(date) {
...
+  }
+
+  i_pkg_deps <- pkgdepends::new_pkg_deps(
+    ifelse(
+      inherits(i_ref_minver, "remote_ref_github"),
+      i_ref_minver$ref,
+      i_ref$ref
+    ),
+    config = list(dependencies = "hard", cran_mirror = ppm_repo)
+  )
```
pawelru, line 64: We need to use an empty libpath without already installed packages

```suggestion
  config <- append_config(default_config(), list(dependencies = "hard", cran_mirror = ppm_repo))

  i_pkg_deps <- pkgdepends::new_pkg_deps(
    ifelse(
      inherits(i_ref_minver, "remote_ref_github"),
      i_ref_minver$ref,
      i_ref$ref
    ),
    config = config
  )
```

R/utils.R
```diff
@@ -31,3 +34,37 @@ get_ppm_snapshot_by_date <- function(date) {
...
+    ppm_repo <- parse_ppm_url(get_ppm_snapshot_by_date(i_release_date))
+  }
+
+  i_pkg_deps <- pkgdepends::new_pkg_deps(
+    ifelse(
+      inherits(i_ref_minver, "remote_ref_github"),
+      i_ref_minver$ref,
+      i_ref$ref
+    ),
+    config = list(dependencies = "hard", cran_mirror = ppm_repo)
```
pawelru, line 63:
```suggestion
    config = list(dependencies = "hard", cran_mirror = ppm_repo, library = tempfile())
```

R/check.R
```diff
@@ -135,48 +133,26 @@ solve_ip.min_isolated_deps_installation_proposal <- function(ip) { # nolint
...
-      i_ref <- pkgdepends::parse_pkg_ref(i_ref_str)
+    if (i_pkg %in% base_pkgs()) return(NULL)
 
-      i_ref_minver <- get_ref_min_incl_cran(i_ref, i_op, i_op_ver)
+    resolve_ppm_snapshot(deps[i, "ref"], deps[i, "op"], deps[i, "version"])
+  })
 
-      i_release_date <- get_release_date(i_ref_minver)
+  new_res <- do.call(rbind, deps_res)
+  new_res <- new_res[!duplicated(new_res[, c("ref", "package", "version")]), ]
```
pawelru, line 152:
```suggestion
  new_res <- do.call(
    rbind,
    lapply(
      split(new_res, new_res$ref),
      function(i) {
        max_ver <- max(package_version(i$version))
        i[package_version(i$version) == max_ver, ]
      }
    )
  )
  new_res <- new_res[!duplicated(new_res), ]
```

R/check.R
```diff
@@ -135,48 +133,26 @@ solve_ip.min_isolated_deps_installation_proposal <- function(ip) { # nolint
...
-      i_res <- i_pkg_deps$get_resolution()
-      i_res$direct <- i_res$directpkg <- FALSE
-      i_res
-    }
-  )
-
-  new_res <- rbind(res[1, ], do.call(rbind, deps_res))
-  new_res <- new_res[!duplicated(new_res), ]
+  # Keep res at top
+  new_res <- rbind(res[1, ], new_res)
```
pawelru, line 155:
```suggestion
  new_res <- rbind(res[1:2, ], new_res)
```

pawelru (approved):

PR #27, opened 2023-07-13
Fix tests by using `solve_ip(x)`

averissimo:
### Motivation

I was thinking on a way to test some of the changes I've been making and realized that tests were outdated.

None of the tests were passing, I corrected them and simplified the code to remove repetitive lines that are difficult to support.

_note_: this PR is not included in the other one to avoid having 1 big PR.

### Current status

Not all of the tests currently pass (pre or post-simplification), but it might be the `Remotes` bug you mentioned or something else that I need to correct.

Regardless these tests should be correct.

### Changes description

* Tests was very outdated with `x$solve()` being used instead of `solve_ip()`
* Adds `local_description()` helper
  * Uses `{withr::local_*}` logic to create a temporary DESCRIPTION file and then remove from disk
  * Supports adding an arbitrary number of imports
  * Supports adding an arbitrary number of remotes
* Adds `test_proposal_common()` helper
  * Avoids having repetitive code for what is essentially the same testing logic.
  * The function has a bit of complexity but cleans up the tests very well
* Style: adds an ASCII separator between test sections
  * [figlet](http://www.figlet.org/) uses a modified figlet script

@pawelru would you prefer to have your branch as the base?

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -28,11 +28,14 @@ skip_if_empty_gh_token <- function() {
 #' the key
 #' @param remotes (`vector`) string vector that contains remotes to add to
 #' the DESCRIPTION file
+#' @param need_verdepcheck (`vector`) string vector that contains
+#' Config/Need/verdepcheck elements to add to the DESCRIPTION file
 #' @param .local_envir (`envirnoment`) The environment to use for scoping.
 #'
 #' @keywords internal
 local_description <- function(pkg_list = c(pkgdepends = "Import"),
                               remotes = c(),
+                              need_verdepcheck = c(),
                               .local_envir = parent.frame()) {
 
   d_std <- desc::desc("!new")
@@ -43,6 +46,10 @@ local_description <- function(pkg_list = c(pkgdepends = "Import"),
 
   for (remote in remotes) {
     d_std$add_remotes(remote)
+  }
+
+  if (!is.null(verdepcheck) && length(verdepcheck) > 0) {
+    d_std$set("Config/Needs/verdepcheck", paste(need_verdepcheck, collapse = ","))
   }
 
   path <- tempfile(pattern = "DESCRIPTION")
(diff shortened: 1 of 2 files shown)


pawelru:
> @pawelru would you prefer to have your branch as the base?

Yes please do change the base.

I am aware that the tests are not so great atm but wanted to do this after all packages will be covered. This is much more important.

tests/testthat/helper.R
```diff
@@ -15,3 +18,103 @@ skip_if_empty_gh_token <- function() {
...
+  for (pkg in names(pkg_list)) {
+    d_std$set_dep(pkg, pkg_list[[pkg]])
+  }
+
+  for (remote in remotes) {
+    d_std$add_remotes(remote)
+  }
+
+  if (!is.null(verdepcheck) && length(verdepcheck) > 0) {
+    d_std$set("Config/Needs/verdepcheck", paste(need_verdepcheck, collapse = ","))
```
pawelru, line 52:
```suggestion
    d_std$set(.desc_field, paste(need_verdepcheck, collapse = ", "))
```

tests/testthat/helper.R
```diff
@@ -15,3 +18,103 @@ skip_if_empty_gh_token <- function() {
...
+
+  for (pkg in names(pkg_list)) {
+    d_std$set_dep(pkg, pkg_list[[pkg]])
+  }
+
+  for (remote in remotes) {
+    d_std$add_remotes(remote)
+  }
+
+  if (!is.null(verdepcheck) && length(verdepcheck) > 0) {
```
pawelru, line 51: it seems that this object (i.e. `verdepcheck` is not defined)

averissimo:
Rebased against training and corrected the variable name problem you identified! Thanks

This was a low-effort sidequest that I won't continue for now

However, the tools/fixes here will allow for unit tests of some of the changes I've been adding

pawelru (approved):
LGTM 👍 Thanks!

PR #28, opened 2023-08-01
Remotes cleanup refactoring, code cleanup & bug fixing

averissimo:
##  Overview of changes description

note: I've added some comments in the code below for additional context.

* `desc_remotes_cleanup()` refactoring just as we discussed
  * I couldn't remove `Remotes` while keeping functionality
* Adds tests
* Code reorganization
  * Moved `solve_*` functions to own file
  * Moved some functions around _(especially to `desc_util.R`, for description-based functions)_
  * `local_description()` moved to `R/`
    * Useful function to create better examples for developers
* when resolving packages: Replaces `remote_github_ref` to CRAN when resolving packages if that version (rule) exists on CRAN
  * For example: `formatters (>= 0.5.1)` isn't on CRAN, so it keeps the github_ref (instead of replacing with non-existent CRAN versions)
* Fix Bioconductor lack of published date bug _([see comment below](https://github.com/insightsengineering/verdepcheck/pull/28/files/bb9188253793357d5006f1ec09287c60b5a94372#diff-d0a3f8001b5b5d8dcd4919cb061472e35e4d41c41a84a3b7c30222f84147d08d))_
  * `parse_ppm_url()` fallback to lastest from PPM if snapshot date cannot be resolved
* Adds utility `map_key_character` to avoid repetitive `vapply()`


tests/testthat/helper.R
```diff
@@ -21,45 +21,6 @@ skip_if_empty_gh_token <- function() {
...
-    d_std$set(.desc_field, paste(need_verdepcheck, collapse = ", "))
-  }
-
-  path <- tempfile(pattern = "DESCRIPTION")
-  d_std$write(path)
-  withr::defer(unlink(path), envir = .local_envir)
-
-  path
-}
-
```
averissimo, line 62: Moved to `R/` as it is recommended by `{testthat}` and is a valuable function for examples.

R/check.R
```diff
@@ -97,156 +97,6 @@ execute_ip <- function(ip, path, build_args, check_args, ...) {
...
-  }
-
-  ip$.__enclos_env__$private$plan$.__enclos_env__$private$resolution$result <- resolution
-
-  ip$solve()
-
-  return(invisible(ip))
-}
-
-
```
averissimo, line 249: Moved to `solve.R` for cleaner structure of files

R/get_ref.R
```diff
@@ -51,8 +51,17 @@ get_ref_min_incl_cran.remote_ref_github <- function(remote_ref, op = "", op_ver
...
+#' check_if_on_cran(list(package = "magrittr"))
+#' check_if_on_cran(list(package = "magrittr"), list(op = ">=", op_ver = "0.5.0"))
+#' check_if_on_cran(list(package = "magrittr"), list(op = ">=", op_ver = "9999.9.99"))
+#' check_if_on_cran(list(package = "magrittr"), list(op = "<", op_ver = "0.0.0"))
+check_if_on_cran <- function(remote_ref, version = NULL) {
+  cran_listings <- pkgcache::meta_cache_list(remote_ref$package)
+  if (is.null(version) || NROW(cran_listings) == 0) return(NROW(cran_listings) > 0)
+  # Check if minimum version exists on CRAN
+  NROW(filter_valid_version(cran_listings$version, version$op, version$op_ver)) > 0
 }
```
averissimo, line 65: Adds `version` parameter to avoid turning a remote GitHub reference to a CRAN if that particular version doesn't exist on CRAN.

This is useful for a scenario where the dependency has been bumped up, but that version has not been released yet on CRAN.

R/get_ref.R
```diff
@@ -250,16 +264,6 @@ get_desc_from_gh <- function(org, repo, ref = "") {
...
-#' @keywords internal
-filter_valid_version <- function(x, op, op_ver) {
-  res <- x
-  res <- Filter(Negate(is.na), res)
-  if (op != "" && op_ver != "") {
-    res <- Filter(function(x) do.call(op, list(x, package_version(op_ver))), res)
-  }
-  return(res)
-}
-
```
averissimo, line 262: Moved to `desc_utils.R`

R/get_ref.R
```diff
@@ -414,11 +418,62 @@ get_release_date.remote_ref <- function(remote_ref) {
...
+  error = function(err) {
+    cli::cli_alert_info(paste0(
+      "Couldn't find the release date for ",
+      cli::col_blue("bioc::{{{package}}}"),
+      ". This may cause a failure to install packages as",
+      " min_cohort / min_isolated strategies depend on the release date."
+    ))
+    NA
+  })
+}
```
averissimo, line 450: Problem with Bioconductor versions as they don't have any metadata information about the publication date.

```R
r> pkgcache::meta_cache_list("SummarizedExperiment")$published
[1] NA
```

**This is very open to change!!**

Workaround checks the last modified date of the package source.

However, I couldn't find an API that would retrieve the information and using Bioconductor release date doesn't work as specific edge case has 2 packages with new releases _(where 1 depends on the other)_.

R/utils.R
```diff
@@ -22,16 +22,26 @@ base_pkgs <- function() {
...
+    ))
+    return(NA)
+  }
   res
 }
 
 #' @importFrom pkgcache ppm_repo_url
-parse_ppm_url <- function(snapshot) {
+parse_ppm_url <- function(snapshot = NA) {
+  if (is.na(snapshot)) return(pkgcache::default_cran_mirror())
```
averissimo, line 44: Fallback to current CRAN in case package published date is `NA`  _(this applies directly in case workaround is reverted)_

R/utils.R
```diff
@@ -67,3 +77,94 @@ resolve_ppm_snapshot <- function(pkg_ref_str, operator, pkg_version) {
...
+#' @keywords internal
+#'
+#' @examples
+#' verdepcheck:::map_key_character(
+#'   list(list(a = "1", b = "2"), list(a = "3", b = "4"), list(a = "5", b = "6")),
+#'   "a"
+#' )
+map_key_character <- function(x, key) {
+  vapply(x, `[[`, character(1), key)
+}
```
averissimo, line 170: Utility to avoid repetitive code

R/utils.R
```diff
@@ -22,16 +22,28 @@ base_pkgs <- function() {
...
+    return(NA)
+  }
   res
 }
 
 #' @importFrom pkgcache ppm_repo_url
-parse_ppm_url <- function(snapshot) {
+parse_ppm_url <- function(snapshot = NA) {
+  if (is.na(snapshot)) {
+    return(pkgcache::default_cran_mirror())
```
averissimo, line 45: Fallback to current CRAN in case package published date is NA (this applies directly in case workaround is reverted)

R/get_ref.R
```diff
@@ -414,11 +420,64 @@ get_release_date.remote_ref <- function(remote_ref) {
...
+      cli::cli_alert_info(paste0(
+        "Couldn't find the release date for ",
+        cli::col_blue("bioc::{{{package}}}"),
+        ". This may cause a failure to install packages as",
+        " min_cohort / min_isolated strategies depend on the release date."
+      ))
+      NA
+    }
+  )
+}
```
averissimo, line 454: Problem with Bioconductor versions as they don't have any metadata information about the publication date.

```R
r> pkgcache::meta_cache_list("SummarizedExperiment")$published
[1] NA
```

**This is very open to change!!**

Workaround checks the last modified date of the package source.

However, I couldn't find an API that would retrieve the information and using Bioconductor release date doesn't work as specific edge case has 2 packages with new releases _(where 1 depends on the other)_.

R/get_ref.R
```diff
@@ -51,8 +51,19 @@ get_ref_min_incl_cran.remote_ref_github <- function(remote_ref, op = "", op_ver
...
+#' verdepcheck:::check_if_on_cran(list(package = "magrittr"), list(op = ">=", op_ver = "0.5.0"))
+#' verdepcheck:::check_if_on_cran(list(package = "magrittr"), list(op = ">=", op_ver = "9999.9.99"))
+#' verdepcheck:::check_if_on_cran(list(package = "magrittr"), list(op = "<", op_ver = "0.0.0"))
+check_if_on_cran <- function(remote_ref, version = NULL) {
+  cran_listings <- pkgcache::meta_cache_list(remote_ref$package)
+  if (is.null(version) || NROW(cran_listings) == 0) {
+    return(NROW(cran_listings) > 0)
+  }
+  # Check if minimum version exists on CRAN
+  NROW(filter_valid_version(cran_listings$version, version$op, version$op_ver)) > 0
```
averissimo, line 66: Adds `version` parameter to avoid turning a remote GitHub reference to a CRAN if that particular version doesn't exist on CRAN.

This is useful for a scenario where the dependency has been bumped up, but that version has not been released yet on CRAN.

R/desc_utils.R
```diff
@@ -0,0 +1,201 @@
...
+#' @examples
+#' d <- desc::desc(cmd = "!new")
+#'
+#' d$set_dep("magrittr", type = "Imports", version = "*")
+#' verdepcheck:::version_from_desc("magrittr", d)
+#'
+#' d$set_dep("magrittr", type = "Imports", version = ">= 1.5")
+#' verdepcheck:::version_from_desc("magrittr", d)
+version_from_desc <- function(pkg_name, desc) {
+  subset_versions <- function(.x) (.x$version[.x$package == pkg_name])[[1]]
```
pawelru, line 143: this function is called once so I doubt it's needed. IMHO it actually overcomplicate stuff rather than simplify

R/desc_utils.R
```diff
@@ -0,0 +1,201 @@
...
+    package = pkg_name,
+    version_str = version,
+    op = "",
+    op_ver = ""
+  )
+  if (version == "*" || trimws(version) == "") {
+    return(result)
+  }
+  split_vec <- strsplit(version, " ")[[1]]
+  result$version_str <- version
```
pawelru, line 155: not needed - you already have it
```suggestion
```

R/desc_utils.R
```diff
@@ -0,0 +1,201 @@
...
+#' )
+#' verdepcheck:::desc_remotes_cleanup(d)
+desc_remotes_cleanup <- function(d) {
+  # Parse the `Config/Needs/verdepcheck` to retrieve references and extract package names
+  desc_field_pkgs <- pkgdepends::parse_pkg_refs(get_desc_field_pkgs(d))
+  desc_field_pkg_names <- map_key_character(desc_field_pkgs, "package")
+
+  # Parse the remotes to retrieve the package names
+  remotes <- pkgdepends::parse_pkg_refs(d$get_remotes())
+  remotes_pkg <- map_key_character(remotes, "package")
```
pawelru, line 81: Can we please have some consistent var naming convention? It will be much easier to read.

For instance here we have `xxx_pkgs` and `yyy` indicating ref as a string and `xxx_pkg_names` and `yyy_pkg` as package name as a string. There is a conflict with `xxx_pkgs` and `yyy_pkg` storing different type of data.

R/desc_utils.R
```diff
@@ -0,0 +1,201 @@
...
+#'
+#' @param x `vector` of valid package versions.
+#' @param op `character(1)` relational operator (`>=`, `==`, ...)
+#' @param op_ver `character(1)` or `package_version(1)` with version to compare
+#' with using a relational operator.
+#'
+#' @keywords internal
+#'
+#' @examples
+#' versions <- paste(1:10, rep("0", 10), sep = ".")
```
pawelru, line 171: a little bit simpler

```suggestion
#' versions <- paste(1:10, 0, sep = ".")
```

R/desc_utils.R
```diff
@@ -0,0 +1,201 @@
...
```
pawelru, line 191:
```suggestion
#' versions <- paste(1:10, 0, sep = ".")
```

R/get_ref.R
```diff
@@ -346,7 +352,7 @@ get_release_date <- function(remote_ref) {
 #' @export
 #' @examplesIf gh::gh_token() != ""
 #' remote_ref <- pkgdepends::parse_pkg_ref("insightsengineering/teal@v0.10.0")
-#' get_release_date.remote_ref_github(remote_ref)
+#' verdepcheck:::get_release_date.remote_ref_github(remote_ref)
```
pawelru, line 355: This is exported

```suggestion
#' get_release_date.remote_ref_github(remote_ref)
```

R/get_ref.R
pawelru: Yeah, packages from Bioconductor are little bit more difficult to handle and I don't have a good solution for that. I just did fallback into `remote_ref` class and return NA and then handle NAs in the following codes.

pawelru: There is a (bioconductor) package called BiocPkgTools ([GH page](github.com/seandavi/BiocPkgTools)) that fetches some metadata. I have read it's source code and it seems that they are parsing URL from the current bioc version (obtained via `BiocManager::version()`) into something like this: https://bioconductor.org/packages/3.17/bioc/VIEWS
They then open url connection, read that file (`read.dcf()`) and reports everything including `"Date/Publication"` field.

R/utils.R
pawelru: There are few bad things happening here:
- `get_ppm_snapshot_by_date` warns that CRAN will be used but this is actually handled by a separate function `parse_ppm_url`. There will be nothing wrong if the latter will be called inside but that's not the case.
- Can we return NULL instead of NA. I kind of dislike NA as it is ambigious.
- From what I recall, "/latest" PPM endpoint was suppose to handle all edge cases (including Bioc). Shall we use cran mirror instead? Right now we have two different fallbacks and I am confused
- I would prefer to handle PPM urls inside PPM related functionalities and fallback into cran mirror outside. Something like: `url <- parse_ppm_snapshot(...) %||% pkgcache::default_cran_mirror()`. IMHO that would be much cleaner.

tests/testthat/test-deps_installation_proposal.R
```diff
@@ -162,3 +216,51 @@ test_that("new_min_cohort_deps_installation_proposal correctly handles \">=\" de
...
+test_that("new_min_isolated_deps_installation_proposal correctly handles tern and rtables", {
+  skip_if_offline()
+  skip_if_empty_gh_token()
+
+  d_std_path <- local_description(
+    list(
+      `tern (>= 0.8.3)` = "Import",
+      `rtables (>= 0.6.1)` = "Import",
+      `formatters (>= 0.5.0)` = "Import"
+    ),
```
pawelru, line 229: standard strings should work as well

```suggestion
    list(
      "tern (>= 0.8.3)" = "Import",
      "rtables (>= 0.6.1)" = "Import",
      "formatters (>= 0.5.0)" = "Import"
    ),
```

tests/testthat/test-deps_installation_proposal.R
```diff
@@ -162,3 +216,51 @@ test_that("new_min_cohort_deps_installation_proposal correctly handles \">=\" de
...
+test_that("new_min_isolated_deps_installation_proposal correctly resolves a different version from specifications", {
+  skip_if_offline()
+  skip_if_empty_gh_token()
+
+  d_std_path <- local_description(
+    list(
+      `tern (>= 0.8.3)` = "Import",
+      `rtables (>= 0.6.0)` = "Import",
+      `formatters (>= 0.4.1)` = "Import"
+    )
```
pawelru, line 258:
```suggestion
    list(
      "tern (>= 0.8.3)" = "Import",
      "rtables (>= 0.6.0)" = "Import",
      "formatters (>= 0.4.1)" = "Import"
    )
```

R/get_ref.R
pawelru: Very good change. Thank you!

R/get_ref.R
pawelru: Please have a look at this: https://code.bioconductor.org/browse/hermes/commits/devel

It looks like Bioc is doing artificial vbump even if package has not changed since the last release resulting in artificial update of the last release date. This algorithm cannot handle Bioc packages in the same way as we handle CRAN and GH packages.

R/utils.R
averissimo: This is a fallback for very recent versions that are after the latest `ppm` snapshot.

That is, releases from the last few days may not have a snapshot on `ppm`, this is a rare edge case, but this was happening with a very fresh release.

R/desc_utils.R
averissimo: Done!

R/get_ref.R
```diff
@@ -236,10 +247,15 @@ get_gh_tags <- function(org, repo, max_date = Sys.Date() + 1, min_date = as.Date
...
-
+#' Get DESCRIPTION from GitHub Repository
+#'
 #' @importFrom desc desc
 #' @importFrom gh gh
 #' @keywords internal
+#'
+#' @examples
+#' verdepcheck:::get_desc_from_gh("tidyverse", "tibble")
+#' verdepcheck:::get_desc_from_gh("insightsengineering", "formatters", "v0.5.0")
```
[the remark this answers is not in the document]
averissimo: Done

R/utils.R
averissimo: Refactored the code:
* Output is clear as it returns the URL for the snapshot
  * `parse_ppm_url` is called inside instead
* Handles fallback inside instead of repeated call when this function is called
* Prefer using a parameter as I was testing if the CRAN is really needed or `latest`
  * Couldn't replicate the edge case that made me use CRAN mirror
    * When it depends on very recent packages _(1 or 2 day old)_
    * Latest ppm seems synced with `pkgcache::meta_cache_list`
    * This can be tested by using packages on top of [this list](https://cran.r-project.org/web/packages/available_packages_by_date.html)

https://github.com/insightsengineering/verdepcheck/blob/0819cbdbce04b16be2fe60f921d8677508d241d1/R/utils.R#L27-L49

R/get_ref.R
averissimo: As we discussed removed the custom to check for the timestamp.

Using the current date as a placeholder for Bioc packages as it fails with `SummarizedExperiment` or do you want for it to fail completely? _(I believe it's used on `teal.data`)_

Created issue upstream on https://github.com/r-lib/pkgcache/issues/99

R/desc_utils.R
averissimo: Sorry, missed that name, it should've been called `remotes_pkg_names`

Using `_refs` and `_names` 

https://github.com/insightsengineering/verdepcheck/blob/0819cbdbce04b16be2fe60f921d8677508d241d1/R/desc_utils.R#L80-L90

R/desc_utils.R
```diff
@@ -0,0 +1,208 @@
...
+#'       "r-lib/pkgdepends@*release"
+#'     ),
+#'     need_verdepcheck = c(
+#'       "dplyr",
+#'       "tibble=tidyverse/tibble@v3.2.1"
+#'     )
+#'   )
+#' )
+#' verdepcheck:::desc_remotes_cleanup(d)
+desc_remotes_cleanup <- function(d) {
```
averissimo, line 76: @pawelru 

We still need to use `Remotes` for situations when there's a release on GitHub that doesn't exist on CRAN _(and that version is restricted on the `DESCRIPTION` file)_

#### Context:

When a package has a GitHub release that doesn't exist on CRAN and if we remove the Remotes field, then the `ip$solve()` call will try to resolve 2 refs of the same package `cran::<pkg>` and `Config/Needs/verdepcheck::<pkg>` _(and expects those 2 resolutions to succeed)_

* The first will resolve to the CRAN version and the second to GitHub.
* 👉 The issue appears when the version on CRAN doesn't comply with `DESCRIPTION` restriction.
* For a solution, we would need to add even more code for custom resolution _(effectively mocking `Remotes`)_

This can be seen currently with `formatters (>= 0.5.1)` needed for many of the `insightsengineering` packages _(see example below)_.

It works without the `Remotes` field for the `max` strategy because there's a valid CRAN/PPM release that satisfies the `DESCRIPTION`, it doesn't on edge cases like the `formatters`

```R
test_that("remotes is needed when release is on github, but not on CRAN", {
  path_desc <- local_description(
    pkg_list = list("formatters (>= 0.5.1)" = "Imports"),
    need_verdepcheck = "insightsengineering/formatters"
  )

  result <- min_isolated_deps_check(path_desc)

  expect_equal(result$ip$get_solution()$status, "FAILED")

  path_desc_remotes <- local_description(
    pkg_list = list("formatters (>= 0.5.1)" = "Imports"),
    need_verdepcheck = "insightsengineering/formatters",
    remotes = list("insightsengineering/formatters@*release")
  )

  results_remotes <- min_isolated_deps_check(path_desc_remotes)

  expect_equal(results_remotes$ip$get_solution()$status, "OK")
})
```

averissimo: Now that `Remotes` has been removed from insightsengineering packages `verdepcheck` will fail on those packages.

#### What can be done to workaround this?

Whenever a package resolves to a `github_remote_ref` on the `Config/Needs/verdepcheck` we can artificially add that reference to `Remotes`. 

For example with `min_isolated` strategy, `teal.data` is currently failing on verdepcheck:

image

It resolves the `Config/Needs/verdepcheck` entry for `teal.logger` to be `teal.logger=insightsengineering/teal.logger@v0.1.1`

We can look to cases when it doesn't resolve to CRAN and add those to remotes.

image

R/utils.R
```diff
@@ -20,7 +20,16 @@ base_pkgs <- function() {
 }
 
 #' @importFrom pkgcache ppm_snapshots
-get_ppm_snapshot_by_date <- function(date) {
+#' @examplesIf Sys.getenv("R_USER_CACHE_DIR", "") != ""
+#' get_ppm_snapshot_by_date(NA)
+#' get_ppm_snapshot_by_date("2023-08-01")
+#' get_ppm_snapshot_by_date(Sys.Date() + 10)
+get_ppm_snapshot_by_date <- function(date,
+                                     fallback_repo = file.path(pkgcache::ppm_repo_url(), "latest")) {
```
pawelru, line 28: do we ever use a `fallback_repo` in the codebase?
YAGNI?

pawelru: I mean we use a fallback functionality but I guess making a new function parameter out of this is probably too much

R/utils.R
```diff
@@ -20,18 +20,39 @@ base_pkgs <- function() {
...
+  }
+  parse_ppm_url(res)
 }
 
 #' @importFrom pkgcache ppm_repo_url
-parse_ppm_url <- function(snapshot) {
+parse_ppm_url <- function(snapshot = NA) {
+  if (is.na(snapshot)) {
+    return(NULL)
+  }
```
pawelru, line 55: [low priority]
can we somehow get rid of `NA`? It's ambigious and could hold a lot of different information.
I spotted that you were trying to add `-Inf` to max, there is also a fallback functionality right above so it seems that you are going into the direction of fixing inputs to that function which is a nice move btw. This makes me thinking whether we still need this if here.

R/desc_utils.R
pawelru: I have pushed a commit that aims to address this issue by keeping config/needs/verdepcheck row only so that we would be having only one (successful) resolution. I have checked this locally using your branches and it's all good. Let's try to test it more.

R/desc_utils.R
```diff
@@ -0,0 +1,208 @@
...
+#'
+#' @examples
+#' d <- desc::desc(cmd = "!new")
+#'
+#' d$set_dep("magrittr", type = "Imports", version = "*")
+#' verdepcheck:::version_from_desc("magrittr", d)
+#'
+#' d$set_dep("magrittr", type = "Imports", version = ">= 1.5")
+#' verdepcheck:::version_from_desc("magrittr", d)
+version_from_desc <- function(pkg_name, desc) {
```
pawelru, line 149: [low priority] - reorder arguments to make `desc` first; also: all functions above use `d` instead of `desc` - both names are good but we need to stay consistent

R/get_ref.R
```diff
@@ -51,8 +51,19 @@ get_ref_min_incl_cran.remote_ref_github <- function(remote_ref, op = "", op_ver
...
 #' @keywords internal
-check_if_on_cran <- function(remote_ref) {
-  nrow(pkgcache::meta_cache_list(remote_ref$package)) > 0
+#'
+#' @examplesIf Sys.getenv("R_USER_CACHE_DIR", "") != ""
+#' verdepcheck:::check_if_on_cran(list(package = "dplyr"))
+#' verdepcheck:::check_if_on_cran(list(package = "dplyr"), list(op = ">=", op_ver = "1.1.0"))
+#' verdepcheck:::check_if_on_cran(list(package = "dplyr"), list(op = ">=", op_ver = "9999.9.99"))
+#' verdepcheck:::check_if_on_cran(list(package = "dplyr"), list(op = "<", op_ver = "0.0.0"))
+check_if_on_cran <- function(remote_ref, version = NULL) {
```
pawelru, line 60: `get_ref_min` (and probably other) have `version` argument "unpacked" into two separate arguments `op` and `op_ver` -> please make it more consistent

R/get_ref.R
```diff
@@ -51,8 +51,19 @@ get_ref_min_incl_cran.remote_ref_github <- function(remote_ref, op = "", op_ver
...
+  cran_listings <- pkgcache::meta_cache_list(remote_ref$package)
+  if (is.null(version) || NROW(cran_listings) == 0) {
+    return(NROW(cran_listings) > 0)
+  }
```
pawelru, line 64: condition checked twice

```suggestion
  if (is.null(version)) {
    return(NROW(cran_listings) > 0)
  }
```

R/get_ref.R
```diff
@@ -104,7 +115,7 @@ get_ref_min.remote_ref_cran <- function(remote_ref, op = "", op_ver = "") {
           sep = " ",
           "Possible problem finding release for:",
```
pawelru, line 116:
```suggestion
          "Problem with finding CRAN release meeting following criteria:",
```

R/desc_utils.R
averissimo: The tests I've done with the current `branch` look good. _(current branch as in: that artificially populates `Remotes` with all `verdepcheck` github dependencies)_... The test suite is quite extensive and caught a few issues.

When attempting not to use `Remotes` at all _(with a change on `desc_remotes_cleanup()`)_ it hits a couple of problems

We go back to how `{pkgdepends}` doesn't reconcile packages dependency and version with `Config/Needs/verdepcheck`.

<details>
<summary>⊞ How to test? (click here to expand)</summary>

1. Apply diff below
2. run locally against [`teal.slice@verdepcheck_action`](https://github.com/insightsengineering/teal.slice/tree/verdepcheck_action)

```diff

diff --git a/R/desc_utils.R b/R/desc_utils.R
index 924f1a7..67340d3 100644
--- a/R/desc_utils.R
+++ b/R/desc_utils.R
@@ -94,7 +94,7 @@ desc_remotes_cleanup <- function(d) {
 
   # Create new list of references that will be used as "Remotes"
   new_remotes <- c(
-    map_key_character(desc_field_refs[desc_field_include_ix], "ref"),
+    # map_key_character(desc_field_refs[desc_field_include_ix], "ref"),
     map_key_character(remotes_refs[remotes_include_ix], "ref")
   )
 ```
</details>

#### Q: How flexible are you on populating `Remotes` with `Config/Needs/verdepcheck` versions?

As it is right now, `Remotes` will only be populated with the resolved `Config/Needs/verdepcheck` tags

- for `max` strategy it will be the GH repo name
- for all others it will be specific tags from GH repo
  - It won't use `org/repo@*release` (it will overwrite it if there's a corresponding `Config/Needs/verdepcheck`

File: R/utils.R (hunk 21 above)
Comment (averissimo) [reply R/utils.R]: Done in 7ce8e09

File: R/desc_utils.R (hunk 23 above)
Comment (averissimo) [reply R/desc_utils.R]: done in 95f7989

File: R/get_ref.R (hunk 24 above)
Comment (averissimo) [reply R/get_ref.R]: Unpacked!

File: DESCRIPTION [hunk 27]
```diff
@@ -4,7 +4,9 @@ Title: Check Package using Various Versions of Dependencies
 Version: 0.0.0.9000
 Date: 2023-02-14
 Authors@R:
-    person("Pawel Rucki", , , "pawel.rucki@roche.com", role = c("aut", "cre"))
+    person("Pawel Rucki", , , "pawel.rucki@roche.com", role = c("aut", "cre")),
+    person("Andre Verissimo", , , "andre.verissimo@roche.com", role = "aut"),
```
pawelru, line 8: @averissimo I have added you to the authors to recognise your contributions so far 🥇

pawelru (approved):

pawelru:
This is awesome. Thank you! I have approved this. Let's merge it to the training branch and then the training to the main. I will let you decide when you want to pull the trigger.

PR #31, opened 2023-08-17
Workflow propagation

walkowif:
Propagation of:
* https://github.com/insightsengineering/idr-tasks/issues/597,
* https://github.com/insightsengineering/idr-tasks/issues/620,
* https://github.com/insightsengineering/idr-tasks/issues/621,
* https://github.com/insightsengineering/idr-tasks/issues/629.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.yaml
+++ b/.github/ISSUE_TEMPLATE/release.yaml
@@ -1,0 +1,107 @@
+---
+name: 🚀 Release
+description: Template for package release
+title: "[Release]: <version>"
+labels: ["release"]
+assignees:
+  - KlaudiaBB
+  - cicdguy
+body:
+  - type: markdown
+    attributes:
+      value: |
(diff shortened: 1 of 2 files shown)


pawelru:
Hey @walkowif from the list of task you have mentioned - only one of them is addressed by changes you propose. What about the rest of them?

walkowif:
I copied the list of all tasks as description to all propagation PRs for simplicity 🙂 
For this repository:
* block-list is not extended, because it's not used here in the first place - I only extend it for repositories where it's already being used. However, I can of course add the `enforce-note-blocklist` and `note-blocklist` inputs here if needed.
* release template is added in the PR
* WORDLIST reduction is added with the removal of `spellcheck` workflow condition
* 'Replace `deploy_to_branch` ...' task doesn't need propagation here since the relevant code is not present in the workflow

pawelru:
Thanks. I was actually thinking on the first one - block list. OK, looks like you have considered this already.

pawelru (approved):

PR #42, opened 2024-04-19
implement an alternative way of getting release date

pawelru:
`pkgcache` stopped reporting publish date used as a release date used in min_cohort strategy. As a result, NA is used which makes max release date not correct.
I have reported this (linked issue) and the maintainer agreed it's an issue that needs a fix. Until this happen, please see this temporary workaround of scrapping it from CRAN webpage.

@Melkiades FYI: that's the root cause of the `formatters` issue we have observed


R/get_ref.R
```diff
@@ -438,6 +438,25 @@ get_cran_data <- function(package) {
...
+      ),
+      5
+    )
+    cran_current <- data.frame(
+      type = "cran",
+      package = package,
+      version = scrapped_table[scrapped_table$X1 == "Version:", "X2"][[1]],
+      published = scrapped_table[scrapped_table$X1 == "Published:", "X2"][[1]]
+    )
+  }
```
averissimo, line 459: Should we keep a cache subset of `tools::CRAN_package_db()` with package, version and published data and access that instead?

Given the comment on https://github.com/r-lib/pkgcache/issues/109#issuecomment-2066467347

> ```R
> db <- tools::CRAN_package_db()
> db$`Date/Publication`
> ```

R/get_ref.R
```diff
@@ -438,6 +438,16 @@ get_cran_data <- function(package) {
   cran_current <- pkgcache::meta_cache_list(packages = package)[, c(
     "type", "package", "version", "published"
   )]
+  if (all(is.na(cran_current$published))) {
+    # workaround of https://github.com/r-lib/pkgcache/issues/109
+    db <- subset(tools::CRAN_package_db(), Package == package)
```
averissimo, line 443: This might be very slow as every run of the db call will download the whole `data.frame`.

How about we create a cache inside the package, as calling `tools::CRAN_package_db` every time will be very time consumingAnd initialize 

Add this outside the function (either on the same file or `utils.R`)
    - This sort of follows the same pattern as `pkgcache` in storing the cache on an internal `env`.

```r
pkgenv <- new.env(parent = emptyenv())
```

```suggestion
    if (is.null(pkgenv$cache_db)) {
      pkgenv$cache_db <- tools::CRAN_package_db()
    }
    db <- subset(pkgenv$cache_db, Package == package)

```

pawelru: thanks for a good suggestion. I have applied this

averissimo (approved):
LGTM!

Good thing you applied my original suggestion and not the edit :facepalm:

PR #49, opened 2024-06-20
Add extra

pawelru:


R/desc_utils.R
```diff
@@ -122,6 +122,82 @@ desc_cond_set_refs <- function(d, refs) {
...
+#' @examples
+#' d <- desc::desc(cmd = "!new")
+#' d <- verdepcheck:::desc_add_extra_deps(d, "foo (>= 1.2.3)")
+#' d <- verdepcheck:::desc_add_extra_deps(d, "bar (>= 2.3.4); baz (>= 3.4.5)")
+#' d$get_deps()
+desc_add_extra_deps <- function(d, x) {
+  if (length(x)) {
+    for (x_i in trimws(strsplit(x, "\\;")[[1]])) {
+      x_i_deparsed <- deparse_dep_str(x_i)
+      d$set_dep(x_i_deparsed$package, "Imports", x_i_deparsed$ver_str)
```
averissimo, line 147: I think we should place a safeguard here to fail if we're trying to overwrite a direct dependency

```suggestion
      x_i_deparsed <- deparse_dep_str(x_i)
      if (any(d$get_deps()$package == x_i_deparsed$package)) {
        stop("Cannot add extra_deps with a package that is already defined in DESCRIPTION")
      }
      d$set_dep(x_i_deparsed$package, dep_type, x_i_deparsed$ver_str)

```

```r
testthat::test_that("desc_add_extra_deps throws an error when dependency already exist in DESCRIPTION", {
  d <- desc::desc(cmd = "!new")
  d$set_dep("foo", type = "Suggests", ">= 1.0.0")
  
  expect_error(
    verdepcheck:::desc_add_extra_deps(d, "foo (>= 1.1.0)"),
    "package that is already defined in DESCRIPTION"
  )
})
```

averissimo (changes requested):
This is similar to what I had in mind :100: 

I think we should create a safeguard to prevent manipulation of direct dependencies.

We can't protect against all malicious use, but this is an easy one in my opinion.

R/desc_utils.R
pawelru: Good suggestion! Thanks! I will implement it shortly.

pawelru: Added here b90ae5f878978adcf4f3e7c085c63e5adeafa8d7

averissimo (approved):
LGTM! :100:

PR #51, opened 2024-07-24
include refs for indirect deps

pawelru:
this fixes the issue observed when testing on `teal@479_mirai_pr`
`mirai@main` requires dev version of `nanonext`. We recently added `extra_deps` parameter to be able to include this in the Imports but we also need to use `Config/Needs/verdepcheck` entry for these extra dependencies. Right now all additional entries are dropped in `get_refs_from_desc()`. The change is pretty simple - add extra deps before this function call.

### DIFF
--- a/tests/testthat/test-deps_installation_proposal.R
+++ b/tests/testthat/test-deps_installation_proposal.R
@@ -347,7 +347,7 @@ test_that("indirect dependencies in the config field - ignore on default", {
   withr::defer(unlink(x$get_config()$library))
 
   d_new <- desc::desc(gsub("^deps::", "", x$get_refs()))
-  expect_false("r-lib/rlang" %in% d_new$get_field("Config/Needs/verdepcheck"))
+  expect_false("r-lib/rlang" %in% d_new$get_list("Config/Needs/verdepcheck"))
 })
 
 test_that("indirect dependencies in the config field - include on match with `extra_deps`", {
@@ -360,5 +360,5 @@ test_that("indirect dependencies in the config field - include on match with `ex
   withr::defer(unlink(x$get_config()$library))
 
   d_new <- desc::desc(gsub("^deps::", "", x$get_refs()))
-  expect_false("r-lib/rlang" %in% d_new$get_field("Config/Needs/verdepcheck"))
+  expect_true("r-lib/rlang" %in% d_new$get_list("Config/Needs/verdepcheck"))
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-deps_installation_proposal.R
```diff
@@ -334,3 +334,31 @@ test_that("new_max_deps_installation_proposal correctly handles Bioc package", {
...
+  skip_if_offline()
+
+  d_std_path <- local_description(list(dplyr = "Import"), need_verdepcheck = c("tidyverse/dplyr", "r-lib/rlang"))
+
+  x <- new_max_deps_installation_proposal(d_std_path, extra_deps = "rlang")
+
+  withr::defer(unlink(x$get_config()$library))
+
+  d_new <- desc::desc(gsub("^deps::", "", x$get_refs()))
+  expect_false("r-lib/rlang" %in% d_new$get_field("Config/Needs/verdepcheck"))
```
averissimo, line 363: Is this a typo? If we add extra dep it will keep the `Config/Needs/verdepcheck` as it treats it as normal dependency

```suggestion
  expect_true("r-lib/rlang" %in% d_new$get_list("Config/Needs/verdepcheck"))
```

tests/testthat/test-deps_installation_proposal.R
```diff
@@ -334,3 +334,31 @@ test_that("new_max_deps_installation_proposal correctly handles Bioc package", {
...
+  skip_if_offline()
+
+  d_std_path <- local_description(list(dplyr = "Import"), need_verdepcheck = c("tidyverse/dplyr", "r-lib/rlang"))
+
+  x <- new_max_deps_installation_proposal(d_std_path)
+
+  withr::defer(unlink(x$get_config()$library))
+
+  d_new <- desc::desc(gsub("^deps::", "", x$get_refs()))
+  expect_false("r-lib/rlang" %in% d_new$get_field("Config/Needs/verdepcheck"))
```
averissimo, line 350: I believe you want to use `get_list` method here and in the other test, as the `get_field` returns a string

```suggestion
  expect_false("r-lib/rlang" %in% d_new$get_list("Config/Needs/verdepcheck"))
```

averissimo (changes requested):
Nice PR, 2 comments though:

1. The testing method should be updated (~`get_fielt`~ to `get_list`)
2. I believe the logic on the second is not right

tests/testthat/test-deps_installation_proposal.R
pawelru: yes of course. Too much copy&paste. Thanks for the spot!

averissimo (approved):
Awesome!

PR #53, opened 2024-08-06
include CRAN refs in get_ref_max()

pawelru:
Fix the error observed for max deps check in `teal`
https://github.com/insightsengineering/teal/actions/runs/10262416165/job/28392047028

The root cause is that `promises` package has higher version [on CRAN](https://github.com/cran/promises/blob/5f6a537c4115cd52ddbce93d5373b01924a678e2/DESCRIPTION#L4) compared to the [GitHub](https://github.com/rstudio/promises/blob/e46a4988856e844c16019ddde63223735f9e065e/DESCRIPTION#L4)

```
r$> pkgdepends::parse_pkg_ref("promises") |> get_version()
[1] ‘1.3.0’

r$> pkgdepends::parse_pkg_ref("rstudio/promises") |> get_version()
[1] ‘1.2.1.9001’
```

I consider this as an edge case but still we need to handle this in the package.

I have modified also `get_ref_release()` to make it similar to the `get_ref_max()`.


averissimo (commented):
Works great, good fix! I do have a comment to avoid having duplicate code.

Given the code is the same, we could just consolidate under one function.

Feel free to cherry pick/merge: https://github.com/insightsengineering/verdepcheck/commit/d72769b95d246d9c6588a0e0caa9969db4751344

We don't even need the `add_gh_releases` parameter unless you see a reason why `max` should not also check Github releases.

pawelru:
Thanks for good comments. It's indeed better to avoid repetition. I did cherry pick and add some other small enhancements. Feel free to check this out

averissimo (approved):
LGTM!

Thanks for spotting that `nrow(cran_current > 0)`

PR #54, opened 2024-08-20
remove workaround; use devel of pkgcache

pawelru:
We have a workaround for a bug that had been fixed on main so let's use this and remove workaround


pawelru:
this had been included in https://github.com/insightsengineering/verdepcheck/pull/56

